PR #15, opened 2020-09-15
Implement export interfaces

nguyen102:
Issue #, if available:

Description of changes:
Implement `initiateExport`, `getExportStatus`, and `cancelExport`. For all three methods, write and read data from DDB as necessary. 

For `initiateExport` only implement writing to `DDB`. Invoking stepFunction is a  future SIM task, after the SIM tasks for Glue Export step function is completed.

**Related PR**
https://github.com/awslabs/fhir-works-on-aws-deployment/pull/104

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,7 +20,7 @@ module.exports = {
         'no-empty-function': 'off',
         '@typescript-eslint/no-empty-function': 'error',
         'import/prefer-default-export': 'off',
-        "import/no-extraneous-dependencies": ["error", {"devDependencies": ["**/*.test.ts"]}]
+        'import/no-extraneous-dependencies': ['error', {'devDependencies': ['**/*.test.ts']}]
     },
     settings: {
         'import/resolver': {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,7 +19,8 @@ module.exports = {
         '@typescript-eslint/no-useless-constructor': 'error',
         'no-empty-function': 'off',
         '@typescript-eslint/no-empty-function': 'error',
-        'import/prefer-default-export': 'off'
+        'import/prefer-default-export': 'off',
+        "import/no-extraneous-dependencies": ["error", {"devDependencies": ["**/*.test.ts"]}]
     },
     settings: {
         'import/resolver': {
--- a/src/dataServices/dynamoDbDataService.test.ts
+++ b/src/dataServices/dynamoDbDataService.test.ts
@@ -9,22 +9,33 @@ import AWSMock from 'aws-sdk-mock';
 import { QueryInput } from 'aws-sdk/clients/dynamodb';
 import * as AWS from 'aws-sdk';
 // eslint-disable-next-line import/no-extraneous-dependencies
-import { BundleResponse, BatchReadWriteResponse } from 'fhir-works-on-aws-interface';
+import {
+    BundleResponse,
+    BatchReadWriteResponse,
+    InitiateExportRequest,
+    ResourceNotFoundError,
+} from 'fhir-works-on-aws-interface';
+import { TooManyConcurrentExportRequestsError } from 'fhir-works-on-aws-interface/lib/errors/TooManyConcurrentExportRequestsError';
+import { UnauthorizedAccessError } from 'fhir-works-on-aws-interface/lib/errors/UnauthorizedAccessError';
+import each from 'jest-each';
 import { utcTimeRegExp } from '../../testUtilities/regExpressions';
 import { DynamoDbBundleService } from './dynamoDbBundleService';
 import { DynamoDbDataService } from './dynamoDbDataService';
 import { DynamoDBConverter } from './dynamoDb';
+import DynamoDbParamBuilder from './dynamoDbParamBuilder';
 
 AWSMock.setSDKInstance(AWS);
 
 // eslint-disable-next-line import/order
 import sinon = require('sinon');
 
+beforeEach(() => {
+    expect.hasAssertions();
+});
+afterEach(() => {
+    AWSMock.restore();
+});
 describe('updateResource', () => {
-    afterEach(() => {
-        AWSMock.restore();
-    });
-
     test('Successfully update resource', async () => {
         // BUILD
         const id = '8cafa46d-08b4-4ee4-b51b-803e20ae8126';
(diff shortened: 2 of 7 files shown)


.eslintrc.js
```diff
@@ -19,7 +19,8 @@ module.exports = {
         '@typescript-eslint/no-useless-constructor': 'error',
         'no-empty-function': 'off',
         '@typescript-eslint/no-empty-function': 'error',
-        'import/prefer-default-export': 'off'
+        'import/prefer-default-export': 'off',
+        "import/no-extraneous-dependencies": ["error", {"devDependencies": ["**/*.test.ts"]}]
```
rsmayda: May want to format this file you are using " while other lines are '

Also just to confirm what does this line do?

src/dataServices/dynamoDbDataService.test.ts
```diff
@@ -86,3 +96,173 @@ describe('updateResource', () => {
...
+            callback(null, {});
+        });
+
+        const dynamoDbDataService = new DynamoDbDataService(new AWS.DynamoDB());
+
+        // OPERATE
+        const jobId = await dynamoDbDataService.initiateExport(initiateExportRequest);
+
+        // CHECK
+        expect(jobId).not.toBeUndefined();
```
rsmayda: whew double negatives are hard could we just use:
```suggestion
        expect(jobId).toBeDefined();

```

src/dataServices/dynamoDbDataService.test.ts
```diff
@@ -86,3 +96,173 @@ describe('updateResource', () => {
...
+        try {
+            await dynamoDbDataService.initiateExport(initiateExportRequest);
+        } catch (e) {
+            // CHECK
+            expect(e).toMatchObject(new TooManyConcurrentExportRequestsError());
+        }
+    });
+});
+
+describe('cancelExport', () => {
```
rsmayda, line 208: What if you cancel a job that is already completed?

src/dataServices/dynamoDbDataService.ts
```diff
@@ -139,19 +145,88 @@ export class DynamoDbDataService implements Persistence {
...
+        // const stepFunctionArn = await StartStepFunctionAndGetStepFunctionArn
+
+        const params = DynamoDbParamBuilder.buildPutCreateExportRequest(jobId, initiateExportRequest, '');
+        await this.dynamoDb.putItem(params).promise();
+        return jobId;
+    }
+
+    async throttleExportRequestsIfNeeded(requesterUserId: string) {
+        const MAXIMUM_SYSTEM_LEVEL_CONCURRENT_REQUESTS = 2;
+        const MAXIMUM_CONCURRENT_REQUEST_PER_USER = 1;
```
rsmayda: maybe make them class constants instead of being inside the method?

src/dataServices/dynamoDbDataService.ts
```diff
@@ -139,19 +145,88 @@ export class DynamoDbDataService implements Persistence {
...
+
+        const queryJobStatusParam = DynamoDbParamBuilder.buildQueryExportRequestJobStatus('in-progress');
```
rsmayda: should we account for the 'cancelling' status?

.eslintrc.js
nguyen102: Converting `"` to `'`. 

In test files we would get a lint error for `no-extraneous-dependencies` if the test file imports a library that is included in `devDependencies` instead of `dependencies`. This rule turns off that error for test files.

src/dataServices/dynamoDbDataService.test.ts
nguyen102: updated!

src/dataServices/dynamoDbDataService.test.ts
nguyen102: I'm going to add code to throw an error if we're trying to cancel an export job that is in `completed` or `failed` state as a safety precaution in the persistence layer.

However, technically we should already be able to catch this invalid state in the router section. In the router, we currently pull the job status to check if a user has access to the job. In this check, we can also check that the job is not in the state of `completed` or `failed`. If they're in those state, we can send a 400 error to the user. I'll update the router PR to include this logic.

src/dataServices/dynamoDbDataService.ts
nguyen102: updated

src/dataServices/dynamoDbDataService.ts
nguyen102: Yeah, let's update this to include the `canceling` status. Otherwise a user could have a large number of jobs in the canceling state at a time. 
When a job is in the `canceling` status, the Export job step function should pick up the new state change and start canceling the Glue job. The rate at which this happens depends on how often we want the step function to check for job status. Maybe once every 3 minutes? As soon as the job is in the canceling state, the Glue job should no longer consume DDB read units.  However, for those 3 minutes, the Glue jobs will still be consuming DDB read units.

carvantes: I think that we should check the job cancelling status a lot more often than 3 min. Maybe 10 seconds. 

Checking the job status is cheap(just a ddb query). Export jobs will likely consume thousands of RCU, it's more efficient to poll for status frequently to cancel them ASAP.

nguyen102: Yep, we can do a shorter time. The timing for how often the job status isn't defined in this code base. But for our step function we can set the job status check time to be shorter.

src/dataServices/dynamoDbDataService.ts
```diff
@@ -139,19 +151,105 @@ export class DynamoDbDataService implements Persistence {
...
+            ) {
+                throw new TooManyConcurrentExportRequestsError();
+            }
+        }
+    }
+
+    async getJobsWithExportStatuses(jobStatuses: ExportJobStatus[]): Promise<ItemList> {
+        const jobStatusPromises = jobStatuses.map((jobStatus: ExportJobStatus) => {
+            const queryJobStatusParam = DynamoDbParamBuilder.buildQueryExportRequestJobStatus(jobStatus);
+            return this.dynamoDb.query(queryJobStatusParam).promise();
```
nguyen102, line 191: Doing two queries to get jobs in the state `canceling` and `in-progress` because query only support equality check on partition key.
 https://stackoverflow.com/questions/40316502/aws-dynamodb-querying-all-hashes-in-array

I looked at `BatchGetItems` but that API doesn't support indexes.
https://stackoverflow.com/questions/42501001/how-to-batchget-index-table-in-dynamodb

src/dataServices/dynamoDbParamBuilder.ts
```diff
@@ -102,4 +108,68 @@ export default class DynamoDbParamBuilder {
...
+            jobStatus: 'in-progress',
+            jobFailedMessage: '',
+        };
+        return {
+            TableName: EXPORT_REQUEST_TABLE,
+            Item: DynamoDBConverter.marshall(item),
+        };
+    }
+
+    static buildQueryExportRequestJobStatus(jobStatus: ExportJobStatus) {
```
rsmayda: nit/nice to have: Can further optimize by adding projections; in our case we only want requestorId and job status

https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/mainline/src/dataServices/dynamoDbParamBuilder.ts#L67

src/dataServices/dynamoDbDataService.ts
```diff
@@ -139,19 +151,105 @@ export class DynamoDbDataService implements Persistence {
...
+        // TODO: Start Export Job Step Function
+        // const stepFunctionArn = await StartStepFunctionAndGetStepFunctionArn
+
+        const params = DynamoDbParamBuilder.buildPutCreateExportRequest(jobId, initiateExportRequest, '');
+        await this.dynamoDb.putItem(params).promise();
+        return jobId;
+    }
+
+    async throttleExportRequestsIfNeeded(requesterUserId: string) {
+        const jobStatuses: ExportJobStatus[] = ['canceling', 'in-progress'];
```
rsmayda:
```suggestion
        const jobStatusesToThottle: ExportJobStatus[] = ['canceling', 'in-progress'];

```

nguyen102: updated

src/dataServices/dynamoDbDataService.test.ts
rsmayda: We should only be able to cancel 'in-progress' states right?

src/dataServices/dynamoDbParamBuilder.ts
nguyen102: added

src/dataServices/dynamoDbDataService.test.ts
rsmayda: Already handled

rsmayda (approved):

PR #21, opened 2020-10-01
feat: API to communicate with Integration Transform

nguyen102:
**SIM:**
FHIR-366

**Branching:**
I created an `api` feature branch. I'll create PRs against that branch until the Integration Transform feature is complete. At that point we can use the code in `api` branch to request Open Source approval for a new repository. Then we can put that code into a `fhir-works-on-aws-persistence-api` repo. 

**Description of changes:**

In this first PR, we're just implementing the API to communicate with the Integration Transform.

A future PR, will handle Auth.

In this PR, I deleted all of the code associated with DDB. I then created an `apiDataService.ts` file and an associated `apiDataService.test.ts` file. 


src/dataServices/apiDataService.ts
```diff
@@ -0,0 +1,123 @@
...
+import axios from 'axios';
+
+export class ApiDataService implements Persistence {
+    updateCreateSupported: boolean = false;
+
+    // TODO: Pull this value from AWS Param Store
+    static readonly INTEGRATION_TRANSFORM_URL = 'http://localhost:4000';
+
+    async createResource(request: CreateResourceRequest): Promise<GenericResponse> {
+        let genericResponse: GenericResponse = { message: '', resource: {} };
```
carvantes: Refactor to use `const` instead of `let`. I don't think mutation is warranted here

Also, is `{ message: '', resource: {} }` ever a valid response?

src/dataServices/apiDataService.ts
```diff
@@ -0,0 +1,123 @@
...
+        try {
+            const url = `${ApiDataService.INTEGRATION_TRANSFORM_URL}/${request.resourceType}`;
+            const response = await axios.post(url, request.resource);
+            genericResponse = { message: '', resource: response.data.resource };
+        } catch (e) {
+            console.error(e);
+            this.handleError(e);
+        }
+        return Promise.resolve(genericResponse);
```
carvantes: just `return genericResponse `

async methods wrap the return value in a promise. There's no point in "double-promise-wrapping" the return value

carvantes (changes requested):

src/dataServices/apiDataService.ts
nguyen102: Initially I used a constant for `genericResponse`. However, if I don't return a `genericResponse `at the end of the method eslint will give me the error `Function lacks ending return statement and return type does not include 'undefined'.`

It was because my `handleError` function was throwing an error (as expected) but eslint didn't know that. I think I can refactor this so `handleError` becomes `getError`. Then I can throw that error in the parent method and eslint will be happy.

src/dataServices/apiDataService.ts
nguyen102: sounds good

src/dataServices/apiDataService.ts
```diff
@@ -0,0 +1,116 @@
...
+    UpdateResourceRequest,
+    vReadResourceRequest,
+} from 'fhir-works-on-aws-interface';
+import axios from 'axios';
+
+export class ApiDataService implements Persistence {
+    updateCreateSupported: boolean = false;
+
+    // TODO: Pull this value from AWS Param Store
+    static readonly INTEGRATION_TRANSFORM_URL = 'http://localhost:4000';
```
rsmayda: I would have this value passing in from deployment (have deployment deal with param store)

src/dataServices/apiDataService.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    async createResource(request: CreateResourceRequest): Promise<GenericResponse> {
+        try {
+            const url = `${ApiDataService.INTEGRATION_TRANSFORM_URL}/${request.resourceType}`;
```
rsmayda: These urls should be the SWAGGER api urls right?

src/dataServices/apiDataService.ts
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            await axios.delete(`${ApiDataService.INTEGRATION_TRANSFORM_URL}/${request.resourceType}/${request.id}`);
+            // Don't need to actually return anything to the router
+            return { message: '', resource: {} };
+        } catch (e) {
+            return this.getError(e, request.resourceType, request.id);
+        }
+    }
+
+    getError(e: any, resourceType: string = '', id: string = '') {
```
rsmayda, line 69: when calling axios.post() if that responds in a non-200 does that throw an error?

package.json
```diff
@@ -32,6 +32,7 @@
     "aws-elasticsearch-connector": "^8.2.0",
     "aws-sdk": "^2.610.0",
```
rsmayda: need ES or SDK?

src/dataServices/apiDataService.ts
nguyen102: Yep, I'm using the API definitions here.
https://app.swaggerhub.com/apis/nguyen102/IntegrationFrameworkAPI/2#/

It doesn't include `persistence` in the route like the original API does, because from Integration Transform's point of view, I feel that including `persistence` is extraneous. 
https://app.swaggerhub.com/apis/nguyen102/IntegrationFrameworkAPI/2#/

Although, if Integration Transform does end up supporting search in the future, it would be convenient to have the resource of `persistence` or `search` added to the route. 

@bnurzhanov  what are your thoughts on using 
`POST [IntTran Base Route]/persistence/[resource-type]` to allow for future support of `GET [IntTran Base Route]/search`
vs
`POST [IntTran Base Route]/[resource-type]`

src/dataServices/apiDataService.ts
nguyen102: Yep, non-200 response will result in an error being thrown by axios.

package.json
nguyen102: deleting ES. I was originally leaving SDK to allow for retrieving `IntTran URL` from param store. But since we're now getting the URL in `development` package, we'll no longer need the sdk here.

src/dataServices/apiDataService.ts
bnurzhanov: This value may not be available during deployment. Our use case: user deploys FHIR Works on AWS and then deploys Integration Transform. This value will be know only after user deploys Integration Transform (including Amazon API Gateway)

nguyen102: done

src/dataServices/apiDataService.ts
bnurzhanov: Do we need to implement Integration Transform as RESTful API? The resource type is part of the payload and can be used to transform FHIR resource into payload formatted to comply with requirements of third-party system (for example as HL7v2 message).
I would vote for calling the `/` resource of the Integration Transform API endpoint

nguyen102: I think to keep things consistent, it would be good to have it be RESTful. For example, when creating a resource we can do `POST /`. But to get a resource we would need to do `GET /Patient/[id]`. So instead of having two different resource path, we can keep it consistent by doing HTTP verbs on `/Patient`

bnurzhanov: > `POST [IntTran Base Route]/persistence/[resource-type]` to allow for future support of `GET [IntTran Base Route]/search`
> vs
> `POST [IntTran Base Route]/[resource-type]`

Integration Transforms are stateless and therefore I think it is all just syntactic sugar that does not add additional information to the transform's execution. I am willing to disagree; and commit

bnurzhanov: I see your point; and I agree.

bnurzhanov (approved):

src/dataServices/apiDataService.ts
```diff
@@ -0,0 +1,118 @@
...
+        } catch (e) {
+            throw this.getError(e, request.resourceType, request.id);
+        }
+    }
+
+    async deleteResource(request: DeleteResourceRequest): Promise<GenericResponse> {
+        try {
+            await axios.delete(`${this.INTEGRATION_TRANSFORM_URL}/${request.resourceType}/${request.id}`);
+            // Don't need to actually return anything to the router
+            return { message: '', resource: {} };
```
carvantes: resource is optional. I think it's better to just omit it
```suggestion
            return { message: ''};
```

src/dataServices/apiDataService.ts
```diff
@@ -0,0 +1,118 @@
...
+    UpdateResourceRequest,
+    vReadResourceRequest,
+} from 'fhir-works-on-aws-interface';
+import axios from 'axios';
+
+export class ApiDataService implements Persistence {
+    updateCreateSupported: boolean = false;
+
+    // TODO: Pull this value from AWS Param Store
+    private INTEGRATION_TRANSFORM_URL: string = '';
```
carvantes: empty string is not a valid init value
```suggestion
    private INTEGRATION_TRANSFORM_URL: string;
```

carvantes (changes requested):

src/dataServices/apiDataService.ts
nguyen102: Changing route to include `persistence` in the route.

nguyen102: @rsmayda  since we're now passing in the `INTEGRATION_TRANFORM_URL` into `apiDataService` from `deployment`. This change to use `persistence` is already taken care of.

src/dataServices/apiDataService.ts
nguyen102: updated

src/dataServices/apiDataService.ts
nguyen102: updated

src/dataServices/apiDataService.ts
nguyen102: Updating route to include `/persistence` as a part of the path. The `deployment` package will pass in the base `IntegrationTransform` url and then each method in `apiDataService.ts` can append the proper resource to the base Url. For example `readResource` will append `/persistence` and `vReadResource` will append `/history`

carvantes (approved):

PR #24, opened 2020-10-05
feat: Add Auth token to API requests

nguyen102:
**Issue #, if available:**
FHIR-370

**Description of changes:**

Added `aws4-axio`. This library adds an interceptor to `axios` that enables `IAM Auth`. It uses the current `IAM role` assumed by the `lambda` function to add Auth token to each request.

Integration Transform then uses the Auth token to Authenticate the request.

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -1,0 +1,4 @@
+export default interface Auth {
+    // Hook into axios and intercept requests
+    initialize(): void;
+}
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -1,5 +1,7 @@
+import { AxiosInstance } from 'axios';
+
 export default interface Auth {
     // Hook into axios and intercept requests to add Auth tokens
     // https://masteringjs.io/tutorials/axios/interceptors
-    initialize(): void;
+    attachInterceptor(axiosInstance: AxiosInstance): void;
 }
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -1,4 +1,5 @@
 export default interface Auth {
-    // Hook into axios and intercept requests
+    // Hook into axios and intercept requests to add Auth tokens
+    // https://masteringjs.io/tutorials/axios/interceptors
     initialize(): void;
 }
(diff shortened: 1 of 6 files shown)


src/auth/auth.ts
```diff
@@ -0,0 +1,5 @@
+export default interface Auth {
+    // Hook into axios and intercept requests to add Auth tokens
+    // https://masteringjs.io/tutorials/axios/interceptors
+    initialize(): void;
+}
```
carvantes: This interface doesn't really communicate it's intent. It's just a method with no parameters and no return value.

Also, using interceptors on the "global" axios object is messy since it's hard to trace what interceptors are attached.

I think it'd be better to use an [axios instance](https://github.com/axios/axios#creating-an-instance) instead(also has other benefits like being able to specify a baseURL)

Then you can have an interface method like:
```js
attachInterceptor(axiosInstance: AxiosInstance): void
```

carvantes (changes requested):

src/auth/auth.ts
nguyen102: updated

src/dataServices/apiDataService.ts
```diff
@@ -12,21 +12,23 @@ import {
...
 
 export class ApiDataService implements Persistence {
     updateCreateSupported: boolean = false;
 
-    private INTEGRATION_TRANSFORM_URL: string;
+    private axiosInstance: AxiosInstance;
 
-    constructor(integrationTransformUrl: string) {
-        this.INTEGRATION_TRANSFORM_URL = integrationTransformUrl;
+    constructor(integrationTransformUrl: string, axiosInstance: AxiosInstance | undefined = undefined) {
```
carvantes:
```suggestion
    constructor(integrationTransformUrl: string, axiosInstance?: AxiosInstance = undefined) {
```
double check before merging, but I think these are equivalent

carvantes (approved):

src/dataServices/apiDataService.ts
nguyen102: that worked.

PR #25, opened 2020-10-08
fix: Auth requires IntTran region

nguyen102:
Description of changes:
* Add config for getting IntTran region
* Update docs with information about IntTran

Note: Changes will need to be made to `package.json` too. For now I'm leaving those changes [here](https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/api-package-json/package.json), until we get approval for the repo `fhir-works-on-aws-persistence-facade`

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -1,3 +1,8 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
 import { AxiosInstance } from 'axios';
 
 export default interface Auth {
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -8,5 +8,9 @@ import { AxiosInstance } from 'axios';
 export default interface Auth {
     // Hook into axios and intercept requests to add Auth tokens
     // https://masteringjs.io/tutorials/axios/interceptors
-    attachInterceptor(axiosInstance: AxiosInstance): void;
+    attachInterceptor(axiosInstance: AxiosInstance, interceptorConfig: InterceptorConfig): void;
 }
+
+export type InterceptorConfig = {
+    awsRegion?: string;
+};
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -8,9 +8,5 @@ import { AxiosInstance } from 'axios';
 export default interface Auth {
     // Hook into axios and intercept requests to add Auth tokens
     // https://masteringjs.io/tutorials/axios/interceptors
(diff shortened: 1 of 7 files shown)


src/auth/auth.ts
```diff
@@ -1,7 +1,16 @@
...
 import { AxiosInstance } from 'axios';
 
 export default interface Auth {
     // Hook into axios and intercept requests to add Auth tokens
     // https://masteringjs.io/tutorials/axios/interceptors
-    attachInterceptor(axiosInstance: AxiosInstance): void;
+    attachInterceptor(axiosInstance: AxiosInstance, interceptorConfig: InterceptorConfig): void;
 }
+
+export type InterceptorConfig = {
```
nguyen102: I added this type, because for the `Auth` interface, we might need other config values in the future. For example, we might need to pass an `api key` to an `Auth` implementation in the future.

src/auth/auth.ts
```diff
@@ -1,7 +1,16 @@
...
+    awsRegion?: string;
+};
```
carvantes: I think that the `InterceptorConfig` type is not helping.

```ts
// ts valid code that fails(I think) at runtime.
IamAuth().attachInterceptor(axiosInstance, {})
```
Is `awsRegion` truly optional? What happens if you pass undefined as region to `aws4Interceptor`? 
Also, this means that **every** Auth implementation accepts awsRegion as a config option. Is that what we want?

carvantes (changes requested):
.

src/auth/auth.ts
nguyen102: I was trying to find a way to allow the `Auth` interface to handle different use cases. For example, how to handle an `Auth` implementation that only requires a `region`. How to handle an `Auth` implementation that only requires an `api-key`. 

After further thought, I think the best way to do that is not through the interface, but through the `constructor` of the class implementing `Auth`. In this case `IamAuth` requires a region, so we'll pass the region to the class through the constructor. Other classes that implement `Auth` can pass in whichever params they require through their own constructor.

src/auth/iamAuth.ts
```diff
@@ -1,13 +1,23 @@
...
 
-const { AWS_REGION } = process.env;
 export default class IamAuth implements Auth {
+    region: string;
+
+    constructor(region: string) {
+        this.region = region;
+    }
+
     // eslint-disable-next-line class-methods-use-this
```
carvantes: This eslint exception can now be removed

carvantes (approved):

src/auth/auth.ts
carvantes: Much better

src/auth/iamAuth.ts
nguyen102: removed

PR #27, opened 2020-10-15
feat: add lambda fns to start glue jobs and check its status

carvantes:
Description of changes:

Added 2 more lambdas that'll be part of the bulk export state machine:
- **startExportJob**: Starts the Glue etl job.
- **getJobStatus**: Gets the Glue job status and also checks if the bulk export request has been canceled by the user. The state machine will use this info to determine the next step.

Also added types to standardize the inputs/outputs of all the step functions lambdas.

Some minor changes may be needed later on once we have the actual Glue etl script.

Related PRs:
- https://github.com/awslabs/fhir-works-on-aws-deployment/pull/122

### DIFF
--- a/src/bulkExport/getJobStatus.test.ts
+++ b/src/bulkExport/getJobStatus.test.ts
@@ -1,0 +1,150 @@
+import * as AWSMock from 'aws-sdk-mock';
+import AWS from 'aws-sdk';
+import { QueryInput } from 'aws-sdk/clients/dynamodb';
+import { getJobStatusHandler } from './getJobStatus';
+import { BulkExportStateMachineGlobalParameters } from './types';
+import { DynamoDBConverter } from '../dataServices/dynamoDb';
+
+AWSMock.setSDKInstance(AWS);
+
+describe('getJobStatus', () => {
+    beforeEach(() => {
+        process.env.GLUE_JOB_NAME = 'jobName';
+        AWSMock.restore();
+    });
+
+    test('completed job', async () => {
+        const event: BulkExportStateMachineGlobalParameters = {
+            jobId: '1',
+            exportType: 'system',
+            transactionTime: '',
+            executionParameters: {
+                glueJobRunId: 'jr_1',
+            },
+        };
+        process.env.GLUE_JOB_NAME = 'jobName';
+        AWSMock.mock('Glue', 'getJobRun', (params: any, callback: Function) => {
+            callback(null, {
+                JobRun: {
+                    JobRunState: 'SUCCEEDED',
+                },
+            });
+        });
+        AWSMock.mock('DynamoDB', 'getItem', (params: QueryInput, callback: Function) => {
+            callback(null, {
+                Item: DynamoDBConverter.marshall({
+                    jobId: '2a937fe2-8bb1-442b-b9be-434c94f30e15',
+                    jobStatus: 'in-progress',
+                }),
+            });
+        });
+        await expect(getJobStatusHandler(event, null as any, null as any)).resolves.toEqual({
+            jobId: '1',
+            exportType: 'system',
+            transactionTime: '',
+            executionParameters: {
+                glueJobRunId: 'jr_1',
+                glueJobRunStatus: 'SUCCEEDED',
+                isCanceled: false,
+            },
+        });
+    });
+
+    test('failed job', async () => {
+        const event: BulkExportStateMachineGlobalParameters = {
+            jobId: '1',
+            exportType: 'system',
+            transactionTime: '',
+            executionParameters: {
+                glueJobRunId: 'jr_1',
+            },
+        };
+        AWSMock.mock('Glue', 'getJobRun', (params: any, callback: Function) => {
+            callback(null, {
+                JobRun: {
+                    JobRunState: 'FAILED',
+                },
+            });
+        });
+        AWSMock.mock('DynamoDB', 'getItem', (params: QueryInput, callback: Function) => {
+            callback(null, {
+                Item: DynamoDBConverter.marshall({
+                    jobId: '2a937fe2-8bb1-442b-b9be-434c94f30e15',
+                    jobStatus: 'in-progress',
+                }),
+            });
+        });
+        await expect(getJobStatusHandler(event, null as any, null as any)).resolves.toEqual({
+            jobId: '1',
+            exportType: 'system',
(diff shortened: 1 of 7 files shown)


src/bulkExport/getJobStatus.test.ts
```diff
@@ -0,0 +1,150 @@
+import * as AWSMock from 'aws-sdk-mock';
```
nguyen102, line 6: Do we need the attribution comment for this file?
```
/*
 *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 *  SPDX-License-Identifier: Apache-2.0
 */
```

src/bulkExport/types.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Bulk export state machine parameters.
+ * All lambda functions in the state machine are expected to use this type as both input and output
+ */
+export interface BulkExportStateMachineGlobalParameters {
+    jobId: string;
+    exportType: ExportType;
+    transactionTime: string;
+    requestQueryParams?: {
```
nguyen102: I think we can just move `_outputFormat`, `_since`, and `_type` to the root level of this interface.We don't need to nest it in `requestQueryParams`. In `export-request` table we store these value at the root level. Also in the interface it's the same.
https://github.com/awslabs/fhir-works-on-aws-interface/blob/bulk-export/src/bulkDataAccess.ts#L5

src/bulkExport/startExportJob.ts
```diff
@@ -0,0 +1,27 @@
...
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Handler } from 'aws-lambda';
+import AWS from 'aws-sdk';
+import { BulkExportStateMachineGlobalParameters } from './types';
+
+export const startExportJobHandler: Handler<
+    BulkExportStateMachineGlobalParameters,
```
nguyen102, line 11: Just checking, what params does `Handler` take? Are the two params supposed to be `BulkExportStateMachineGlobalParameters` and `BulkExportStateMachineGlobalParameters` again?

carvantes: Handler is a type that represents a lambda handler function. In this case `BulkExportStateMachineGlobalParameters` is used as both the input event type and the result type.
```js
export type Handler<TEvent = any, TResult = any> = (
    event: TEvent,
    context: Context,
    callback: Callback<TResult>,
) => void | Promise<TResult>;
```

src/bulkExport/getJobStatus.test.ts
carvantes: Added.

We should do something that checks at build time that all the proper copyright headers are in place

src/bulkExport/types.ts
carvantes: updated

src/bulkExport/getJobStatus.ts
```diff
@@ -0,0 +1,47 @@
...
+import { BulkExportStateMachineGlobalParameters } from './types';
+import DynamoDbParamBuilder from '../dataServices/dynamoDbParamBuilder';
+import { DynamoDBConverter } from '../dataServices/dynamoDb';
+
+export const getJobStatusHandler: Handler<
+    BulkExportStateMachineGlobalParameters,
+    BulkExportStateMachineGlobalParameters
+> = async event => {
+    const { GLUE_JOB_NAME } = process.env;
+    if (GLUE_JOB_NAME === undefined) {
```
rsmayda, line 17: Do we want to use falsely here instead of undefined? https://stackoverflow.com/questions/3982663/falsey-values-in-javascript

src/bulkExport/getJobStatus.ts
```diff
@@ -0,0 +1,47 @@
...
+        throw new Error('GLUE_JOB_NAME environment variable is not defined');
+    }
+    const glueJobRunId = event.executionParameters?.glueJobRunId;
+    if (glueJobRunId === undefined) {
```
rsmayda, line 21: Use falsely?

src/bulkExport/getJobStatus.ts
carvantes: I'm ok with just doing an undefined check here. There's no meaningful difference with a falsy check here.

This lambda will fail anyways due to the call to Glue if there's an invalid `GLUE_JOB_NAME` and that's ok. That'd mean that something is wrong on the CFN template and a redeployment is needed to fix it.

This is just a convenience to have an explicit error if env vars are missing since that's the most likely dev error when assembling the step functions state machine

src/bulkExport/getJobStatus.ts
carvantes: Similar as above. I'm ok with just doing an undefined check here. There's no meaningful difference with a falsy check here.

This lambda will fail anyways due to the call to Glue if the `glueJobRunId` is invalid and that's ok. That'd mean that precedent steps in the state machine are generating invalid/incomplete outputs and need to be fixed.

This is just a convenience to have an explicit error if outputs from previous steps are missing since that's the most likely dev error when assembling the step functions state machine.

src/bulkExport/startExportJob.ts
```diff
@@ -0,0 +1,27 @@
...
+    BulkExportStateMachineGlobalParameters
+> = async event => {
+    const { GLUE_JOB_NAME } = process.env;
+    if (GLUE_JOB_NAME === undefined) {
+        throw new Error('GLUE_JOB_NAME environment variable is not defined');
+    }
+    const glue = new AWS.Glue();
+    const startJobRunResponse = await glue.startJobRun({ JobName: GLUE_JOB_NAME }).promise();
```
nguyen102: Can we provide a `jobId` to `startJobRun`? This can be used by the job script to place the exported files in the correct bucket.
For example all exported files for job `abcd1234` would go into bucket 
`export/abcd1234`

rsmayda (commented):
Looks good to me I'd like Tim to give final approval, as I've been out of the loop

src/bulkExport/startExportJob.ts
carvantes: updated to pass arguments `jobID` among other arguments to startJobRun

nguyen102 (approved):

PR #28, opened 2020-10-19
feat: add bulk export updateStatus lambda

carvantes:
Description of changes:

Add the `updateStatus` lambda handler. This lambda function simply updates the status in DDB for a given jobId. The new status will be specified in the state machine definition.

Related PRs:
- https://github.com/awslabs/fhir-works-on-aws-deployment/pull/124

### DIFF
--- a/src/bulkExport/updateStatus.test.ts
+++ b/src/bulkExport/updateStatus.test.ts
@@ -5,6 +5,7 @@
 
 import * as AWSMock from 'aws-sdk-mock';
 import AWS from 'aws-sdk';
+import each from 'jest-each';
 import { updateStatusStatusHandler } from './updateStatus';
 
 AWSMock.setSDKInstance(AWS);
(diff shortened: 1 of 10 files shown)


src/bulkExport/updateStatus.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Handler } from 'aws-lambda';
+import AWS from 'aws-sdk';
+import { ExportJobStatus } from 'fhir-works-on-aws-interface';
+import DynamoDbParamBuilder from '../dataServices/dynamoDbParamBuilder';
+
+const EXPORT_JOB_STATUS = ['completed', 'failed', 'in-progress', 'canceled', 'canceling'];
+const isJobStatus = (x: string): x is ExportJobStatus => EXPORT_JOB_STATUS.includes(x);
+
+export const updateStatusStatusHandler: Handler<{ jobId: string; status: string }, void> = async event => {
+    const { jobId, status } = event;
```
rsmayda, line 15: Need to validate jobId at all? Or is this Update also a create, or is that in the DDB parameter builder?

src/bulkExport/updateStatus.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    test('valid status', async () => {
+        AWSMock.mock('DynamoDB', 'updateItem', (params: any, callback: Function) => {
+            callback(null);
+        });
+        await expect(
+            updateStatusStatusHandler({ jobId: '1', status: 'completed' }, null as any, null as any),
+        ).resolves.toBeUndefined();
+    });
+
+    test('invalid status', async () => {
```
rsmayda: Possible to have a null status?

carvantes: null/undefined are invalid status. I just added tests for them

src/bulkExport/updateStatus.ts
carvantes: It should be on the DDB param builder, but it is not. 
I just updated `DynamoDbParamBuilder` to update only existing jobs. In general We could use some refactoring around the ddb interactions. Relying on `DynamoDbParamBuilder` alone is not a great abstraction.

nguyen102 (approved):

PR #29, opened 2020-10-26
feat: add s3 presigned urls to bulk export status response

carvantes:
Issue #, if available:

Description of changes:

Export results are stored in with the following name format: `<jobId>/<Resource-Type>-<file #>.ndjson`

This lists the objects for a given jobId, generates presigned urls and returns them as part of the `$export/<jobId>` response.

Note: I'm not a fan of the `const SOME_ENV_VAR = process.env.SOME_ENV_VAR || ''` approach. Ideally this package shouldn't be aware of env variables and just receive all those vars as function/constructor params. However, this package consumes env variables in many places and doing such refactor is out of scope for this PR.

Related PRs:
- https://github.com/awslabs/fhir-works-on-aws-deployment/pull/127

### DIFF
--- a/src/bulkExport/bulkExportResults.test.ts
+++ b/src/bulkExport/bulkExportResults.test.ts
@@ -3,7 +3,7 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 import * as AWSMock from 'aws-sdk-mock';
-import AWS from 'aws-sdk';
+import AWS from '../AWS';
 import { getBulkExportResults } from './bulkExportResults';
 
 AWSMock.setSDKInstance(AWS);
--- a/src/bulkExport/bulkExportResults.test.ts
+++ b/src/bulkExport/bulkExportResults.test.ts
@@ -1,0 +1,61 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+import * as AWSMock from 'aws-sdk-mock';
+import AWS from 'aws-sdk';
+import { getBulkExportResults } from './bulkExportResults';
+
+AWSMock.setSDKInstance(AWS);
+
+describe('getBulkExportResults', () => {
+    beforeEach(() => {
+        process.env.GLUE_JOB_NAME = 'jobName';
+        AWSMock.restore();
+
+        AWSMock.mock('STS', 'assumeRole', (params: any, callback: Function) => {
+            callback(null, {
+                Credentials: { AccessKeyId: 'xxx', SecretAccessKey: 'xxx', SessionToken: 'xxx' },
+            });
+        });
+
+        AWSMock.mock('S3', 'getSignedUrl', (apiCallToSign: any, params: any, callback: Function) => {
+            callback(null, 'https://somePresignedUrl');
+        });
+    });
+
+    test('happy case', async () => {
+        AWSMock.mock('S3', 'listObjectsV2', (params: any, callback: Function) => {
+            callback(null, {
+                Contents: [{ Key: 'job-1/Patient-1.ndjson' }, { Key: 'job-1/Observation-1.ndjson' }],
+            });
+        });
+
+        await expect(getBulkExportResults('job-1')).resolves.toEqual([
+            { type: 'Patient', url: 'https://somePresignedUrl' },
+            { type: 'Observation', url: 'https://somePresignedUrl' },
+        ]);
+    });
+
+    test('no results', async () => {
+        AWSMock.mock('S3', 'listObjectsV2', (params: any, callback: Function) => {
+            callback(null, {
+                Contents: [],
+            });
+        });
+
+        await expect(getBulkExportResults('job-1')).resolves.toEqual([]);
+    });
+
+    test('filenames with unknown format', async () => {
+        AWSMock.mock('S3', 'listObjectsV2', (params: any, callback: Function) => {
+            callback(null, {
+                Contents: [{ Key: 'job-1/BadFilenameFormat$$.exe' }, { Key: 'job-1/Observation-1.ndjson' }],
+            });
+        });
+
+        await expect(getBulkExportResults('job-1')).rejects.toThrowError(
+            'Could not parse the name of bulk exports result file: job-1/BadFilenameFormat$$.exe',
+        );
+    });
+});
--- a/src/bulkExport/bulkExportResults.ts
+++ b/src/bulkExport/bulkExportResults.ts
@@ -49,7 +49,7 @@ const signExportResults = async (keys: string[]): Promise<{ key: string; url: st
 };
 
 const getResourceType = (key: string, jobId: string): string => {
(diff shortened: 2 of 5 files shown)


src/bulkExport/bulkExportResults.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    test('filenames with unknown format', async () => {
+        AWSMock.mock('S3', 'listObjectsV2', (params: any, callback: Function) => {
+            callback(null, {
+                Contents: [{ Key: 'job-1/BadFilenameFormat$$.exe' }, { Key: 'job-1/Observation-1.ndjson' }],
+            });
+        });
+
+        await expect(getBulkExportResults('job-1')).rejects.toThrowError(
+            'Could not parse the name of bulk exports result file: job-1/BadFilenameFormat$$.exe',
+        );
```
rsmayda, line 59: What output is returned to the customer in the error case?

carvantes: It's unhandled. It'll end up as a 5xx error. 
There's no way to recover from this. It'd mean that the glue ETL job is generating bad files.

src/bulkExport/bulkExportResults.ts
```diff
@@ -0,0 +1,67 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+import AWS from 'aws-sdk';
+
+const EXPIRATION_TIME_SECONDS = 1800;
+const EXPORT_RESULTS_BUCKET = process.env.EXPORT_RESULTS_BUCKET || ' ';
+const EXPORT_RESULTS_SIGNER_ROLE_ARN = process.env.EXPORT_RESULTS_SIGNER_ROLE_ARN || '';
```
rsmayda, line 9: If you weren't going to use env variables would you just make this a class instead?

src/bulkExport/bulkExportResults.ts
```diff
@@ -0,0 +1,67 @@
...
+                Bucket: EXPORT_RESULTS_BUCKET,
+                Key: key,
+                Expires: EXPIRATION_TIME_SECONDS,
+            }),
+        })),
+    );
+};
+
+const getResourceType = (key: string, jobId: string): string => {
+    const regex = new RegExp(`^${jobId}/([A-Za-z]+)-.*$`);
```
rsmayda: I am not sure how the files are being created but is the `.*` prescriptive enough? In your PR message it appears it may just be a number so perhaps you want to use `\d+`

src/bulkExport/bulkExportResults.ts
```diff
@@ -0,0 +1,67 @@
...
+    const match = regex.exec(key);
```
rsmayda, line 53: When to use `regex.exec(key)` vs `key.match(regex)`? This is some somewhat clear guidance: https://stackoverflow.com/questions/9214754/what-is-the-difference-between-regexp-s-exec-function-and-string-s-match-fun/33139816#:~:text=exec()%20returns%20only%20the,See%20here%3A%20exec%2C%20match.&text=If%20your%20regex%20is%20global,then%20you%20must%20use%20exec.

But curious your thoughts

carvantes: afaik they are interchangeable unless you are doing global matching

> If the regular expression does not include the g flag, str.match() will return the same result as RegExp.exec().

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

src/bulkExport/bulkExportResults.ts
carvantes: Either a class or a function with more params like `getBulkExportResults(bucket, role, jobId)`

src/bulkExport/bulkExportResults.ts
carvantes: I just updated the regex. 

yeah, I left it vague since we weren't 100% sure about how glue would name the files. and I just needed to parse the resourcetype. Now that we know that we are manually renaming the files at the end of the glue job we can be more specific about it.

src/bulkExport/bulkExportResults.ts
```diff
@@ -0,0 +1,67 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+import AWS from 'aws-sdk';
```
nguyen102: Should this import from `../AWS`? I believe that allows us to track AWS calls using XRay?

src/bulkExport/bulkExportResults.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const EXPIRATION_TIME_SECONDS = 1800;
+const EXPORT_RESULTS_BUCKET = process.env.EXPORT_RESULTS_BUCKET || ' ';
+const EXPORT_RESULTS_SIGNER_ROLE_ARN = process.env.EXPORT_RESULTS_SIGNER_ROLE_ARN || '';
+
+const getFiles = async (jobId: string): Promise<string[]> => {
+    const s3 = new AWS.S3();
+
+    const listObjecsResult = await s3.listObjectsV2({ Bucket: EXPORT_RESULTS_BUCKET, Prefix: jobId }).promise();
```
nguyen102: listObjecsResult => listObjectsResult

missing `t`

src/bulkExport/bulkExportResults.ts
nguyen102: Are we not using `../AWS` because we're using the `SIGNER_ROLE_ARN`?

src/bulkExport/bulkExportResults.ts
nguyen102: Can this use `../AWS` and therefore be tracked in XRay?

carvantes: updated

src/bulkExport/bulkExportResults.ts
carvantes: I switched to using `../AWS`. The signer role does not interfere with x ray tracing. It is just a a parameter to a call to STS.

Actually this change is not required to make x-ray tracing work. It looks like 
```js
import AWS from 'aws-sdk';
const AWSWithXray = AWSXRay.captureAWS(AWS);
``` 
actually modifies the `AWS` config/prototype, so you don't really need to pass around the `AWSWithXray` result. Other modules can just `import AWS from 'aws-sdk';` and they'll have x-ray enabled. This is not obvious when looking at the docs of how to enable xray in nodejs.

This means that there's a less intrusive way to enable x-ray for an entire package. Still, updated it to be consistent for now. 

https://github.com/aws/aws-xray-sdk-node/blob/fdbbdf7b7e788a70d23d25637296b254a6171a2f/packages/core/lib/patchers/aws_p.js#L37

rsmayda (approved):

nguyen102 (approved):

PR #31, opened 2020-10-28
feat: add bulk export getCrawlerStatus lambda

carvantes:
Description of changes:

Another lambda function that'll be part of the bulk export state machine. 

This lambda function simply checks the status of the crawler that was started earlier by the `startCrawler` task and reports the status on its response. The state machine will use this status to decide the next step

Related PRs:
- https://github.com/awslabs/fhir-works-on-aws-deployment/pull/130

### DIFF
--- a/src/bulkExport/getCrawlerStatus.test.ts
+++ b/src/bulkExport/getCrawlerStatus.test.ts
@@ -1,0 +1,64 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+import * as AWSMock from 'aws-sdk-mock';
+import AWS from 'aws-sdk';
+import each from 'jest-each';
+import { BulkExportStateMachineGlobalParameters } from './types';
+import { getCrawlerStatusHandler } from './getCrawlerStatus';
+
+AWSMock.setSDKInstance(AWS);
+
+describe('getCrawlerStatusHandler', () => {
+    beforeEach(() => {
+        process.env.CRAWLER_NAME = 'crawlerName';
+        AWSMock.restore();
+    });
+
+    each([
+        ['READY', 'SUCCEEDED', 'succeeded'],
+        ['READY', 'CANCELLED', 'failed'],
+        ['READY', 'FAILED', 'failed'],
+        ['RUNNING', '**any**', 'running'],
+        ['STOPPING', '**any**', 'running'],
+    ]).test('State=%s, LastCrawl.Status=%s => %s', async (state: string, lastStatus: string, expected: string) => {
+        AWSMock.mock('Glue', 'getCrawler', (params: any, callback: Function) => {
+            callback(null, {
+                Crawler: {
+                    LastCrawl: {
+                        Status: lastStatus,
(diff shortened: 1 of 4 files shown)


src/bulkExport/getCrawlerStatus.ts
```diff
@@ -0,0 +1,65 @@
...
+            // This should never happen per the current Glue API specification
+            throw new Error(`Unknown crawler state: ${crawlerState}`);
+    }
+};
+
+export const getCrawlerStatusHandler: Handler<
+    BulkExportStateMachineGlobalParameters,
+    BulkExportStateMachineGlobalParameters
+> = async event => {
+    const { CRAWLER_NAME } = process.env;
```
rsmayda, line 45: Is the crawler name always the same? just curious if this is the best way to pass in the parameter (env var or request param)

src/bulkExport/getCrawlerStatus.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+AWSMock.setSDKInstance(AWS);
+
+describe('getCrawlerStatusHandler', () => {
+    beforeEach(() => {
+        process.env.CRAWLER_NAME = 'crawlerName';
+        AWSMock.restore();
+    });
+
+    each([
```
rsmayda, line 19: We should add the two error states:
- Unknown last crawl status: ${lastCrawlStatus}
- Unknown crawler state: ${crawlerState}

src/bulkExport/getCrawlerStatus.ts
carvantes: The crawler name is always the same. 
It is similar to how DDB table names are always the same and env variables are the preferred way to pass them to Lambda.

src/bulkExport/getCrawlerStatus.test.ts
carvantes: I did not test those errors on purpose.  

Those 2 code branches exist but they are unreachable unless Glue makes a breaking API change, which is extremely unlikely to happen. A test case like "when Glue does a breaking change on their response structure, then getCrawlerStatus should fail with \*error msg\*" isn't really useful.

How does that sound to you?

https://docs.aws.amazon.com/glue/latest/webapi/API_LastCrawlInfo.html#Glue-Type-LastCrawlInfo-Status
https://docs.aws.amazon.com/glue/latest/webapi/API_Crawler.html#Glue-Type-Crawler-State

nguyen102: As a general rule of thumb, I generally add unit tests to cover all of the code path for the "system under test". In this case the "system under test" has a code path for throwing these "unknown" errors.

However, that is a good point that these edge cases are unlikely to be encountered during run time. 

Personally I would just cover the edge case since it's easy to do so, and as a general rule as stated above. However, this is not a blocking change for me.

nguyen102 (approved):

src/bulkExport/getCrawlerStatus.test.ts
rsmayda: Good to know context and the implausibility of the edge cases. I'd prefer to add them but not blocking for me either

PR #32, opened 2020-10-30
Bulk export stop job

nguyen102:
Issue #, if available:

Description of changes:
* Add stop export job
* The next step after this stage is `updateStatusToCanceled` [link](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/f9a4e0522a68de709ccb948a54b2b530593fa582/bulkExport/state-machine-definition.yaml#L34). This step requires the value `jobId` so we're returning that value in the `stopExportJob` stage.

### DIFF
--- a/src/bulkExport/stopExportJob.test.ts
+++ b/src/bulkExport/stopExportJob.test.ts
@@ -1,0 +1,74 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+import * as AWSMock from 'aws-sdk-mock';
+import AWS from 'aws-sdk';
+import { BulkExportStateMachineGlobalParameters } from './types';
+import { stopExportJobHandler } from './stopExportJob';
+
+AWSMock.setSDKInstance(AWS);
+
+describe('getJobStatus', () => {
+    beforeEach(() => {
+        process.env.GLUE_JOB_NAME = 'jobName';
+        AWSMock.restore();
+    });
+
+    test('stop job successfully', async () => {
+        const glueJobRunId = 'jr_1';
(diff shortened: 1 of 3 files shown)


src/bulkExport/stopExportJob.ts
```diff
@@ -0,0 +1,34 @@
...
+    const glue = new AWS.Glue();
+    const stopJobRunResponse = await glue
+        .batchStopJobRun({
+            JobName: GLUE_JOB_NAME,
+            JobRunIds: [glueJobRunId],
+        })
+        .promise();
+    if (stopJobRunResponse.Errors!.length > 0) {
+        console.log('Failed to stop job', JSON.stringify(stopJobRunResponse));
+        throw new Error(`Failed to stop job`);
```
rsmayda: beneficial to add jobid here?

src/bulkExport/stopExportJob.test.ts
```diff
@@ -0,0 +1,74 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+import * as AWSMock from 'aws-sdk-mock';
+import AWS from 'aws-sdk';
+import { BulkExportStateMachineGlobalParameters } from './types';
+import { stopExportJobHandler } from './stopExportJob';
+
+AWSMock.setSDKInstance(AWS);
+
+describe('getJobStatus', () => {
```
rsmayda, line 12: Should we cover the other two error cases? (bad env settings)

src/bulkExport/stopExportJob.ts
nguyen102: added

src/bulkExport/stopExportJob.test.ts
nguyen102: added

carvantes (approved):

PR #35, opened 2020-11-03
feat: start bulk export state machine execution

carvantes:
Start the step functions state machine execution for bulk export requests. 

Note that I removed the `stepFunctionExecutionArn` from DDB. Instead I'm passing the jobId as the execution name.
step functions optionally allows to pass a unique** name when you start the execution and that name is part of the execution ARN( `arn:aws:states:<region>:<account>:execution:<state machine name>:<execution name>`) jobId is unique so it can be used as the execution name. AFAIK the execution arn was stored in ddb only for bookkeeping/debugging and is not used anywhere.

Also did some minor refactoring since I wanted to avoid duplication of the default job values (like outputFormat=ndjson, etc.)

### DIFF
--- a/src/bulkExport/bulkExport.ts
+++ b/src/bulkExport/bulkExport.ts
@@ -3,10 +3,12 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 import AWS from '../AWS';
+import { BulkExportJob } from './types';
 
 const EXPIRATION_TIME_SECONDS = 1800;
 const EXPORT_RESULTS_BUCKET = process.env.EXPORT_RESULTS_BUCKET || ' ';
 const EXPORT_RESULTS_SIGNER_ROLE_ARN = process.env.EXPORT_RESULTS_SIGNER_ROLE_ARN || '';
+const EXPORT_STATE_MACHINE_ARN = process.env.EXPORT_STATE_MACHINE_ARN || '';
 
 const getFiles = async (jobId: string): Promise<string[]> => {
     const s3 = new AWS.S3();
@@ -64,4 +66,23 @@ export const getBulkExportResults = async (jobId: string): Promise<{ type: strin
         type: getResourceType(key, jobId),
         url,
     }));
+};
+
+export const startJobExecution = async (bulkExportJob: BulkExportJob): Promise<void> => {
+    const { jobId, exportType, transactionTime, outputFormat, since } = bulkExportJob;
+    const params = {
+        jobId,
+        exportType,
+        transactionTime,
+        outputFormat,
+        since,
+    };
+
+    await new AWS.StepFunctions()
+        .startExecution({
+            stateMachineArn: EXPORT_STATE_MACHINE_ARN,
+            name: jobId,
+            input: JSON.stringify(params),
+        })
+        .promise();
 };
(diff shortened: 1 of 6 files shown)


src/bulkExport/bulkExport.ts
```diff
@@ -65,3 +67,22 @@ export const getBulkExportResults = async (jobId: string): Promise<{ type: strin
...
+
+export const startJobExecution = async (bulkExportJob: BulkExportJob): Promise<void> => {
+    const { jobId, exportType, transactionTime, outputFormat, since } = bulkExportJob;
+    const params = {
+        jobId,
+        exportType,
+        transactionTime,
+        outputFormat,
+        since,
+    };
```
rsmayda, line 79: No changes: but an interesting Slack overflow on (de)constructing: https://stackoverflow.com/a/44144355/9041033

src/dataServices/dynamoDbDataService.ts
```diff
@@ -35,7 +35,25 @@ import { DynamoDbBundleService } from './dynamoDbBundleService';
...
+import { getBulkExportResults, startJobExecution } from '../bulkExport/bulkExport';
+import { BulkExportJob } from '../bulkExport/types';
+
+const buildExportJob = (initiateExportRequest: InitiateExportRequest): BulkExportJob => {
+    const initialStatus: ExportJobStatus = 'in-progress';
+    return {
+        jobId: uuidv4(),
+        jobOwnerId: initiateExportRequest.requesterUserId,
+        exportType: initiateExportRequest.exportType,
+        groupId: initiateExportRequest.groupId ?? '',
```
rsmayda, line 47: No changes required -- interesting usage I haven't heard of [nullish-coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing)

Any concerns with this being our default on how to default set a parameter?

src/dataServices/dynamoDbDataService.ts
```diff
@@ -35,7 +35,25 @@ import { DynamoDbBundleService } from './dynamoDbBundleService';
...
+        outputFormat: initiateExportRequest.outputFormat ?? 'ndjson',
+        since: initiateExportRequest.since ?? '1800-01-01T00:00:00.000Z', // Default to a long time ago in the past
```
rsmayda, line 49: I am a bit behind on this code flow in general but what does `since` represent? is it the `createdAt` time?

rsmayda (commented):
Approved on my end but I lack context around the removal of stepFunctionExecutionArn from DDB. I would like @nguyen102 give final approval

src/dataServices/dynamoDbDataService.ts
nguyen102: It's a request query param that users can set. `since` is a filter that tells the export job to only pull records from `DDB` that have a `meta.lastUpdated` value that is newer than `since`

nguyen102 (approved):

src/dataServices/dynamoDbDataService.ts
carvantes: yes, using it for assigning default values is maybe it's most popular use case. There aren't any downsides in using it.

PR #44, opened 2020-12-11
feat: Add reference to data model

nguyen102:
Issue #, if available:
FHIR-439

Description of changes:
Add logic to Persistence to add _references, for all references within the object.

**A/C:**
* new updates/creates add a _references list in the DynamoDB object
* read/searches does not return _references in the result

### DIFF
--- a/src/dataServices/dynamoDbBundleService.test.ts
+++ b/src/dataServices/dynamoDbBundleService.test.ts
@@ -9,7 +9,6 @@ import AWSMock from 'aws-sdk-mock';
 import { QueryInput, TransactWriteItemsInput } from 'aws-sdk/clients/dynamodb';
 import * as AWS from 'aws-sdk';
 import { BundleResponse, BatchReadWriteRequest } from 'fhir-works-on-aws-interface';
-import { SinonSandbox } from 'sinon';
 import { DynamoDbBundleService } from './dynamoDbBundleService';
 import { DynamoDBConverter } from './dynamoDb';
 import { timeFromEpochInMsRegExp, utcTimeRegExp, uuidRegExp } from '../../testUtilities/regExpressions';
@@ -21,17 +20,9 @@ import sinon = require('sinon');
 AWSMock.setSDKInstance(AWS);
 
 describe('atomicallyReadWriteResources', () => {
-    let sandbox: SinonSandbox;
-    beforeEach(function() {
-        sandbox = sinon.createSandbox();
-    });
-
-    afterEach(function() {
-        sandbox.restore();
-    });
-
     afterEach(() => {
         AWSMock.restore();
+        sinon.restore();
     });
 
     const id = 'bce8411e-c15e-448c-95dd-69155a837405';
@@ -115,7 +106,7 @@ describe('atomicallyReadWriteResources', () => {
                 });
             });
 
-            const transactWriteItemStub = sandbox.stub();
+            const transactWriteItemStub = sinon.stub();
             // LOCK Items (Success)
             transactWriteItemStub.onFirstCall().returns({ error: null, value: {} });
 
@@ -144,7 +135,7 @@ describe('atomicallyReadWriteResources', () => {
         // When creating a resource, no locks is needed because no items in DDB to put a lock on yet
         async function runCreateTest(shouldReqHasReferences: boolean) {
             // BUILD
-            const transactWriteItemSpy = sandbox.spy();
+            const transactWriteItemSpy = sinon.spy();
             AWSMock.mock('DynamoDB', 'transactWriteItems', (params: TransactWriteItemsInput, callback: Function) => {
                 transactWriteItemSpy(params);
                 callback(null, {});
@@ -268,7 +259,7 @@ describe('atomicallyReadWriteResources', () => {
 
         async function runUpdateTest(shouldReqHasReferences: boolean) {
             // BUILD
-            const transactWriteItemSpy = sandbox.spy();
+            const transactWriteItemSpy = sinon.spy();
             AWSMock.mock('DynamoDB', 'transactWriteItems', (params: TransactWriteItemsInput, callback: Function) => {
                 transactWriteItemSpy(params);
                 callback(null, {});
@@ -296,7 +287,7 @@ describe('atomicallyReadWriteResources', () => {
             }
             const newResource = { ...oldResource, test: 'test' };
 
-            sandbox
+            sinon
                 .stub(DynamoDbHelper.prototype, 'getMostRecentResource')
                 .returns(Promise.resolve({ message: 'Resource found', resource: oldResource }));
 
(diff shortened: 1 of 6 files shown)


src/dataServices/dynamoDbUtil.test.ts
```diff
@@ -50,6 +51,8 @@ describe('cleanItem', () => {
 describe('prepItemForDdbInsert', () => {
     const id = '8cafa46d-08b4-4ee4-b51b-803e20ae8126';
     const vid = 1;
+    const organization = 'Organization/1';
+    const otherPatient = 'Patient/pat2';
```
carvantes: I wouldn't declare things here if they are being used for exactly 1 test. It's just harder to read.

src/dataServices/dynamoDbBundleService.test.ts
```diff
@@ -9,16 +9,27 @@ import AWSMock from 'aws-sdk-mock';
 import { QueryInput, TransactWriteItemsInput } from 'aws-sdk/clients/dynamodb';
 import * as AWS from 'aws-sdk';
 import { BundleResponse, BatchReadWriteRequest } from 'fhir-works-on-aws-interface';
+import { SinonSandbox } from 'sinon';
```
carvantes: Why?

sinon > 5.0.0 defaults to behave as a sandbox, so we may not need to create sandboxes explicitly

https://sinonjs.org/releases/latest/sandbox/

** btw, not sure why we are using sinon instead of Jest mock functions to begin with.

src/dataServices/dynamoDbUtil.test.ts
nguyen102: Ah, let me move it inside the test where it's being used so the context is clearer and it's easier to read. The reason I declared these two strings as variables is because they're each used twice. Once as part of the BUILD step in "BUILD, OPERATE, CHECK", and the second time as part of the CHECK step. It's important that the values are the same in both steps.

src/dataServices/dynamoDbBundleService.test.ts
nguyen102: In this case, if we don't use `sandbox`, we'll get this error. 
`TypeError: Attempted to wrap getMostRecentResource which is already wrapped`

In this test file we call `runUpdateTest` twice. This results in this [method](https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/a27a0c72b7ac57d8d6f592af5dd99cc1bee7e039/src/dataServices/dynamoDbBundleService.test.ts#L299) being called twice and causing the error above. Using sandbox fixes that issue.

As to why we're using `sinon`. In the early stages of the project we started using sinon for `mocks` and `spy` because at the time we didn't know about jest mocks/spy and wasn't familiar with it. Technically we can convert all of the old code using `sinon` to jest, but that's out of scope for this PR. Also that would probably be one of the lower priority tech debt items.

src/dataServices/dynamoDbUtil.ts
```diff
@@ -39,6 +42,19 @@ export class DynamoDbUtil {
...
         item[DOCUMENT_STATUS_FIELD] = documentStatus;
         item[LOCK_END_TS_FIELD] = Date.now();
+
+        // Format of flattenedResource
+        // https://www.npmjs.com/package/flat
+        // flatten({ key1: { keyA: 'valueI' } })  => { key1.keyA: 'valueI'}
+        const flattenedResources: Record<string, string> = flatten(resource);
+        const references = Object.keys(flattenedResources)
+            .filter((key: string) => {
+                return key.includes('reference');
```
carvantes: This is a bug. A field can contain the string "reference" without being a reference i.e. `MolecularSequence.referenceSeq`

https://www.hl7.org/fhir/molecularsequence.html

carvantes (changes requested):

src/dataServices/dynamoDbBundleService.test.ts
```diff
@@ -9,16 +9,27 @@ import AWSMock from 'aws-sdk-mock';
...
+        sandbox = sinon.createSandbox();
+    });
+
+    afterEach(function() {
+        sandbox.restore();
+    });
+
     afterEach(() => {
         AWSMock.restore();
     });
```
carvantes: combine into a single `afterEach`

src/dataServices/dynamoDbBundleService.test.ts
carvantes: The issue is fixed by calling `.restore();` after each test, not by using sandboxes.

You could simply use `sinon.restore()` on the `afterEach`

src/dataServices/dynamoDbUtil.ts
nguyen102: Good catch. I added a fix for it. And added unit test to catch this case.

The fix is to do an exact match on "reference" and to check that "reference" is the last value in the key.

src/dataServices/dynamoDbBundleService.test.ts
nguyen102: done

src/dataServices/dynamoDbBundleService.test.ts
nguyen102: Good catch! That worked. Thanks for the suggestion.

src/dataServices/dynamoDbUtil.ts
```diff
@@ -39,6 +42,19 @@ export class DynamoDbUtil {
...
         item[DOCUMENT_STATUS_FIELD] = documentStatus;
         item[LOCK_END_TS_FIELD] = Date.now();
+
+        // Format of flattenedResource
+        // https://www.npmjs.com/package/flat
+        // flatten({ key1: { keyA: 'valueI' } })  => { key1.keyA: 'valueI'}
+        const flattenedResources: Record<string, string> = flatten(resource);
+        const references = Object.keys(flattenedResources)
+            .filter((key: string) => {
+                return key.split('.').pop() === 'reference';
```
carvantes: you may want to use `.endsWith()` instead

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

carvantes (changes requested):

src/dataServices/dynamoDbUtil.ts
nguyen102: updated

src/dataServices/dynamoDbUtil.ts
```diff
@@ -39,6 +42,19 @@ export class DynamoDbUtil {
...
         item[DOCUMENT_STATUS_FIELD] = documentStatus;
         item[LOCK_END_TS_FIELD] = Date.now();
+
+        // Format of flattenedResource
+        // https://www.npmjs.com/package/flat
+        // flatten({ key1: { keyA: 'valueI' } })  => { key1.keyA: 'valueI'}
+        const flattenedResources: Record<string, string> = flatten(resource);
+        const references = Object.keys(flattenedResources)
+            .filter((key: string) => {
+                return key.endsWith('reference');
```
carvantes: Without the dot it may introduce a bug. Not sure if there's an actual FHIR resource that has a non-reference field that ends with "reference" though.
```suggestion
                return key.endsWith('.reference');

```

carvantes (dismissed):
see my comment regarding `.reference`.

Other than that looks good to me

carvantes (approved):

PR #48, opened 2021-01-25
feat: Change ES mapping for keyword parameters

rsmayda:
Description of changes:
- We need to use `keyword` data type for `_references` data
- Add self as a reference, since resources are self-referencing

### DIFF
--- a/src/dataServices/dynamoDbUtil.ts
+++ b/src/dataServices/dynamoDbUtil.ts
@@ -55,7 +55,7 @@ export class DynamoDbUtil {
                 return flattenedResources[key];
             });
         // Add self to references as the resource is self referencing
-        references.push(`${resource.resourceType}/${id}`);
+        references.push(`${resource.resourceType}/${id}`)
         item[REFERENCES_FIELD] = references;
         return item;
     }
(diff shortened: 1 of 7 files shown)


src/dataServices/dynamoDbUtil.ts
```diff
@@ -47,13 +47,15 @@ export class DynamoDbUtil {
...
-        const references = Object.keys(flattenedResources)
+        const references: string[] = Object.keys(flattenedResources)
             .filter((key: string) => {
                 return key.endsWith('.reference');
             })
             .map((key: string) => {
                 return flattenedResources[key];
             });
+        // Add self to references as the resource is self referencing
+        references.push(`${resource.resourceType}/${id}`);
```
carvantes: Why?

IMO this makes the meaning of `_references` harder to understand. A true self referencing resource would be(not very useful, but it can happen):
```json
{
    "id": "111",
    "resourceType": "Observation",
    "derivedFrom": "Observation/111"
}
```

rsmayda: From a searching/filtering perspective if we do not add `self` as a reference, the SMART filter of 'only look up things that I can see', becomes 'look up things that I am referenced in, excluding my own record'

example: for search `GET Patient/?name=robert` and the filter would be _references == 'Patient/123'

What we could do is add an OR filter for `_id` if resource type== search type

Thoughts?

carvantes: I think it's better to update the query with an OR clause to match `_references` or `id`. That makes the code self explanatory regarding the meaning of clinical scopes.

carvantes (approved):

nguyen102 (approved):

PR #51, opened 2021-02-04
fix: Take resourceType into account when doing Read/VRead/deleteVersionedRes Operation

### LINKED ISSUE #50: Throw error if resource found does not match resourceID
**Defect**

Currently `DynamoDbHelper` does not validate that the `resourceType` of the DynamoDB record retrieved via `getMostRecentResource()` or `getMostRecentValidResource()` matches the `resourceType` passed into those functions. This allows an API client to request resources of a different FHIR resource type through the wrong URLs. 

**Actual**
A Patient record with `id=8a9f93c9-0f37-4a98-bb55-76992774a98b` can be retrieved through the URL `/ExplanationOfBenefits/8a9f93c9-0f37-4a98-bb55-76992774a98b`. 

**Expected**
The above Patient should only be accessible via the URL `/Patient/8a9f93c9-0f37-4a98-bb55-76992774a98b`. Additionally, the URL `/ExplanationOfBenefits/8a9f93c9-0f37-4a98-bb55-76992774a98b`, should return an HTTP 404 - Not Found.

nguyen102:
Issue #, if available:

Description of changes:
FHIR-489 
*  Take `resourceType` into account for operations `readResource`, `deleteVersionedResource`, and `vReadResource`
* Add Logic to the CREATE method to not overwrite an existing resources (Check to see if `id` is already present)


src/dataServices/dynamoDbDataService.ts
```diff
@@ -94,13 +97,21 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
         let item = resource;
         item.resourceType = resourceType;
         item.meta = generateMeta(vid.toString());
 
         const params = DynamoDbParamBuilder.buildPutAvailableItemParam(item, id || uuidv4(), vid);
-        await this.dynamoDb.putItem(params).promise();
+        try {
+            await this.dynamoDb.putItem(params).promise();
+        } catch (e) {
+            if (e.code === 'ConditionalCheckFailedException') {
```
nguyen102, line 110: This is the recommended way to check for `ConditionalCheckFailedException` error.
https://stackoverflow.com/a/55538223/14310364

testUtilities/ConditionalCheckFailedException.ts
```diff
@@ -0,0 +1,11 @@
+export class ConditionalCheckFailedExceptionMock extends Error {
```
nguyen102, line 1: This class is created solely for testing purposes. When we need to return `ConditionalCheckFailedException` in mocks, this class is used.

src/dataServices/dynamoDbDataService.ts
```diff
@@ -94,13 +97,21 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
+                throw new Error(`Unable to create resource with id ${id}, because resource already exists`);
```
rsmayda: In a create we have control of the `id`. Instead of an error can we generate a new uuid and retry?

src/dataServices/dynamoDbHelper.ts
```diff
@@ -17,18 +17,37 @@ export default class DynamoDbHelper {
...
 
-    async getMostRecentResource(
+    private async getMostRecentResources(
         resourceType: string,
         id: string,
         projectionExpression?: string,
-    ): Promise<GenericResponse> {
-        const params = DynamoDbParamBuilder.buildGetResourcesQueryParam(id, 1, projectionExpression);
+    ): Promise<ItemList> {
+        const params = DynamoDbParamBuilder.buildGetResourcesQueryParam(id, resourceType, 2, projectionExpression);
```
rsmayda: can we pass in the `2` as a param; perhaps `size`?

src/dataServices/dynamoDbParamBuilder.ts
```diff
@@ -54,20 +57,29 @@ export default class DynamoDbParamBuilder {
...
+        maxNumberOfVersions: number,
+        projectionExpression?: string,
+    ) {
         const params: any = {
             TableName: RESOURCE_TABLE,
             ScanIndexForward: false,
             Limit: maxNumberOfVersions,
+            FilterExpression: '#r = :resourceType',
             KeyConditionExpression: 'id = :hkey',
+            ExpressionAttributeNames: { '#r': 'resourceType' },
```
rsmayda, line 79: do you want to use quotes here?
```suggestion
            ExpressionAttributeNames: { '#r': resourceType },

```

src/dataServices/dynamoDbHelper.ts
```diff
@@ -38,19 +57,12 @@ export default class DynamoDbHelper {
...
-        }
+        const items = await this.getMostRecentResources(resourceType, id);
         const latestItemDocStatus = items[0][DOCUMENT_STATUS_FIELD];
         if (latestItemDocStatus === DOCUMENT_STATUS.DELETED) {
             throw new ResourceNotFoundError(resourceType, id);
         }
+        let item: any = {};
         // If the latest version of the resource is in PENDING, grab the previous version
         if (latestItemDocStatus === DOCUMENT_STATUS.PENDING && items.length > 1) {
             // eslint-disable-next-line prefer-destructuring
```
rsmayda: Is this `if` block valid? What if the document status === PENDING AND items.length === 1? we would return a PENDING resource as valid

src/dataServices/dynamoDbHelper.ts
```diff
@@ -38,19 +57,12 @@ export default class DynamoDbHelper {
     }
 
     async getMostRecentValidResource(resourceType: string, id: string): Promise<GenericResponse> {
```
rsmayda: A bit retrospective here is there a better term than `valid` here? is not that 'DELETED' resources are invalid. I can't think of another name but if you have a better one let me know

src/dataServices/dynamoDbParamBuilder.ts
```diff
@@ -102,12 +114,25 @@ export default class DynamoDbParamBuilder {
         };
     }
 
-    static buildPutAvailableItemParam(item: any, id: string, vid: number) {
+    /**
+     * Build DDB PUT param to insert a new resource
+     * @param item
+     * @param id
+     * @param vid
```
rsmayda: I would argue put in a comment or delete. item would be a good one to add but I think id & vid are self-explanatory

src/dataServices/dynamoDbParamBuilder.ts
```diff
@@ -102,12 +114,25 @@ export default class DynamoDbParamBuilder {
...
         const newItem = DynamoDbUtil.prepItemForDdbInsert(item, id, vid, DOCUMENT_STATUS.AVAILABLE);
-        return {
+        const param: any = {
             TableName: RESOURCE_TABLE,
             Item: DynamoDBConverter.marshall(newItem),
         };
+
+        if (!allowOverwrite) {
+            param.ConditionExpression = 'attribute_not_exists(id)';
+        }
```
rsmayda: Is this method used exclusively for CREATEs? if not all new PUTs will fail since puts create a new row with a new vid and the SAME id

src/dataServices/dynamoDbDataService.ts
nguyen102: If we have an `id` conflict, it's unlikely that the `id` was a randomly generated `uuid`. The conflicting `id` most likely came from the request. If the `id` did come from the request, I think we should let the requester know that the `id` they provided didn't work rather than generate a random new `id` for them.

src/dataServices/dynamoDbHelper.ts
nguyen102: yup. updated

src/dataServices/dynamoDbParamBuilder.ts
nguyen102: Yep, quote is needed here. In this case we want to map `#r` to the string `resourceType`

Similar to here in relation to `year`
```
var params = {
    TableName : "Movies",
    KeyConditionExpression: "#yr = :yyyy",
    ExpressionAttributeNames:{
        "#yr": "year"
    },
    ExpressionAttributeValues: {
        ":yyyy": 1985
    }
};
```
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GettingStarted.NodeJs.04.html

src/dataServices/dynamoDbHelper.ts
nguyen102: Good catch. I updated the logic to handle that case. The logic now explicitly looks for resources in valid status (Available, Pending_Delete, Locked). If the latest resource is in Pending, and there are more than one resource, then return the second latest resource. 

If none of these case apply, return a `ResourceNotFound` exception.

src/dataServices/dynamoDbParamBuilder.ts
nguyen102: deleted `id` and `vid`. Added a comment for `item`

src/dataServices/dynamoDbParamBuilder.ts
nguyen102: Yep, it's only being used for CREATEs.

It's only being used here
https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/9ab088450754c74bd50f9768d728cd692cb64000/src/dataServices/dynamoDbDataService.ts#L106

src/dataServices/dynamoDbHelper.ts
nguyen102: Thoughts on this name: `getMostRecentUserReadableResource`?

I'm also adding comments to the method to make it clearer.
```
/**
  * @return The most recent resource that has not been deleted and has been committed to the database (i.e. The resource is not in a transitional state)
  */
```

src/dataServices/dynamoDbHelper.ts
```diff
@@ -22,7 +22,13 @@ export default class DynamoDbHelper {
         id: string,
         projectionExpression?: string,
     ): Promise<ItemList> {
-        const params = DynamoDbParamBuilder.buildGetResourcesQueryParam(id, resourceType, 2, projectionExpression);
+        const maxNumberOfVersions = 2;
```
rsmayda: whoops I meant add it as a method param
```suggestion
maxNumberVersionsToGet: number,
projectionExpression?: string,
    ): Promise<ItemList> {

```

src/dataServices/dynamoDbDataService.ts
rsmayda: hmmmmm So this will result in a 500 returned 'InternalServiceError'

If the user didn't provide an id we should return a 'Retryable' but if they did provide an id we should  return a 422 Unprocessable Entity or 400 Bad Request

src/dataServices/dynamoDbParamBuilder.ts
rsmayda: Ahh thanks I didn't see the `name` and `values` sections

src/dataServices/dynamoDbHelper.ts
rsmayda: I like good change

src/dataServices/dynamoDbHelper.ts
nguyen102: Updated

src/dataServices/dynamoDbDataService.ts
nguyen102: Updated

src/dataServices/dynamoDbDataService.ts
```diff
@@ -94,13 +98,26 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
+        try {
+            await this.dynamoDb.putItem(params).promise();
+        } catch (e) {
+            if (e.code === 'ConditionalCheckFailedException') {
+                // It is highly unlikely that the autogenerated id will collide with a preexisting id, but adding that case
+                // to cover all bases
+                const errorMessage = id
+                    ? `Unable to create resource with id ${id}, because resource already exists`
+                    : 'Auto generated id matched an existing id, please try again';
+                throw new InvalidResourceError(errorMessage);
```
nguyen102, line 117: Related PR to handle this error
https://github.com/awslabs/fhir-works-on-aws-routing/pull/52

rsmayda (dismissed):

src/dataServices/dynamoDbDataService.ts
```diff
@@ -94,13 +98,26 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
         const params = DynamoDbParamBuilder.buildPutAvailableItemParam(item, id || uuidv4(), vid);
-        await this.dynamoDb.putItem(params).promise();
+        try {
+            await this.dynamoDb.putItem(params).promise();
+        } catch (e) {
+            if (e.code === 'ConditionalCheckFailedException') {
+                // It is highly unlikely that the autogenerated id will collide with a preexisting id, but adding that case
+                // to cover all bases
+                const errorMessage = id
+                    ? `Unable to create resource with id ${id}, because resource already exists`
```
carvantes: why? afaik we do NOT allow customers to choose their own id when creating resources. Are we changing that?

Not sure if that's by design though, all this code looks like we allow custom ids, but the router never passes an id to this function, so custom ids are not possible right now.

carvantes: Allowing the customer to set an id when creating a resource goes against the FHIE spec: https://www.hl7.org/fhir/http.html#create

nguyen102: Since we don't allow customers to choose their own id, and FHIR specs doesn't allow that, I'll remove the `id` functionality from this method.

nguyen102: Updated

src/dataServices/dynamoDbDataService.ts
```diff
@@ -94,13 +98,26 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
+                    : 'Auto generated id matched an existing id, please try again';
```
carvantes: This `Auto generated id matched an existing id, please try again` msg doesn't look right.

This error is being mapped to a 400, which is not correct. There's nothing wrong with the client request.
This sounds like a 500 error, it is the servers job to do what it takes to generate a unique id when creating resources. 

This is not a message that we want to show to customers. In general catching this exception is not worth it since the probability of uuid collisions is  **extremely** low. you COULD retry with a new uuid, but It's ok to just let it be a 500 error that happens once in a million years.

carvantes (changes requested):

src/dataServices/dynamoDbDataService.ts
rsmayda: It is an Upsert that you can create with an ID: https://www.hl7.org/fhir/http.html#upsert

src/dataServices/dynamoDbDataService.ts
nguyen102: Changing the code to throw an Error, that'll be eventually mapped to a 500 with an 'Internal Server Error'.

Now that we don't allow a user to provide an `id` anymore, if there is a `ConditionalCheckFailedException`, it'll be because the randomly generated id match a preexisting id. Like @carvantes mentioned, the chance of this happening is extremely low.

src/dataServices/dynamoDbDataService.ts
rsmayda: But it looks like it is mapped to a PUT which would be routed in as an updateResource not create

carvantes (approved):

PR #53, opened 2021-02-05
fix: resolve vid and meta attribute mismatch on concurrent Update requests

nguyen102:
Issue #, if available:

Description of changes:
FHIR-491

When an Update request is triggered, instead of generating the `meta` attribute and then passing it to the `transactionService`, let's allow `transactionService` to generate the `meta` attribute.

TransactionService code flow
1. [Update request](https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/4f433d2eacdd81c25bdc6e5a2d5e9ea755a33204/src/dataServices/dynamoDbBundleServiceHelper.ts#L69)
2. [meta attribute is  generated if meta is not already present](https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/4f433d2eacdd81c25bdc6e5a2d5e9ea755a33204/src/dataServices/dynamoDbUtil.ts#L32)

**Testing**
I was able to confirm that the Update error existed before the code change, but was fixed after the code change was implemented.


src/dataServices/dynamoDbDataService.ts
```diff
@@ -149,12 +149,7 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
     async updateResource(request: UpdateResourceRequest) {
         const { resource, resourceType, id } = request;
         const resourceCopy = { ...resource };
-        const getResponse = await this.readResource({ resourceType, id });
-        const currentVId: number = getResponse.resource.meta
-            ? parseInt(getResponse.resource.meta.versionId, 10) || 0
-            : 0;
-
-        resourceCopy.meta = generateMeta((currentVId + 1).toString());
+        await this.readResource({ resourceType, id });
```
rsmayda, line 154: add a comment saying `// Will throw NotFoundError if resource can't be found`

src/dataServices/dynamoDbDataService.ts
```diff
@@ -163,14 +158,13 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
         // Sending the request to `atomicallyReadWriteResources` to take advantage of LOCKING management handled by
         // that method
         const response: BundleResponse = await this.transactionService.transaction({
             requests: [batchRequest],
             startTime: new Date(),
         });
-        item = clone(resource);
+        const item = clone(resource);
         const batchReadWriteEntryResponse = response.batchReadWriteResponses[0];
         item.meta = generateMeta(batchReadWriteEntryResponse.vid, new Date(batchReadWriteEntryResponse.lastModified));
```
rsmayda, line 171: Does this explicitly need to happen or is the `meta`  returned from the bundler if we can remove this part I'd feel better

rsmayda (commented):
How did you replicate this test and how do you verify its gone?

src/dataServices/dynamoDbDataService.ts
```diff
@@ -149,12 +149,7 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
     async updateResource(request: UpdateResourceRequest) {
         const { resource, resourceType, id } = request;
         const resourceCopy = { ...resource };
-        const getResponse = await this.readResource({ resourceType, id });
-        const currentVId: number = getResponse.resource.meta
-            ? parseInt(getResponse.resource.meta.versionId, 10) || 0
-            : 0;
-
-        resourceCopy.meta = generateMeta((currentVId + 1).toString());
```
carvantes, line 157: Where are we incrementing the version number now that we removed this? I can't find it

carvantes:
Do we know how many DDB queries are being made for a single `PUT api/<ResourceType>` call?

I think that the locking mechanism being used is overly complex for a single PUT. That makes it hard to predict how it will behave at scale. We already found this bug, but I'm worried there may be other failure modes.

I'd strongly consider rewriting this using a simpler optimistic locking approach.

src/dataServices/dynamoDbDataService.ts
nguyen102: updated

src/dataServices/dynamoDbDataService.ts
nguyen102: For Update requests the Bundler returns an empty `resource` attribute. However, it does return the fields that are needed to regenerate the `meta` attribute.

The thought being for Updates, as long as an error isn't thrown the resource that was provided to the Bundler is the resource that is written to DDB. Only changes being `vid` and the updated timestamp.

src/dataServices/dynamoDbDataService.ts
nguyen102: The `vid` is being updated here.
https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/4f433d2eacdd81c25bdc6e5a2d5e9ea755a33204/src/dataServices/dynamoDbBundleServiceHelper.ts#L68

nguyen102:
I tested by using [Artillery](https://artillery.io/docs/guides/overview/welcome.html#Stay-in-touch) to replicate 4 concurrent requests against the server for 10 seconds. 

In that time frame 40 requests were sent, with 33 successful (200 response code) and 7 failed (500 response code) requests. However, despite the failed requests, the data in DDB was consistent. The `vid` and `meta `attribute still matched up. After the artillery test, I was able to make Update requests to the same resource using POSTMAN. The request was successful.

In regards to the cause for the 500 errors:
The 500 errors is caused by the lambda function failing to lock the resource for an Update. We need to lock the resource before initiating an Update to ensure that no other resources is updating the resource at the same time.
```
ERROR	Failed to lock TransactionCanceledException: Transaction cancelled, please refer cancellation reasons for specific reasons [TransactionConflict]
```
It makes sense why some of the requests failed to process. We were initiating 4 concurrent update requests to the same resource. Those concurrent requests cannot operate on the same resource at the same time, because only one process can lock the resource at a time.

The same test using the original code, resulted in 2 successful request and 38 failed requests. The data in DDB became inconsistent because the `vid` and `metadata` didn't match up. Once that happened all subsequent requests failed.

nguyen102:
As discuss offline with @carvantes, this Update request can probably be optimized to be more efficient. However, refactoring this code to optimize the Update query will require several more days to implement and confirm that the refactor doesn't introduce any new bug. 
Since we're planning to release SMART soon, it is best to fix this soon so it won't block the SMART release. 

I've created a SIM ticket (FHIR-492) so that we can prioritize the refactor in the future.

To answer the question of the number of DDB requests and Update requires. It is 5 requests. The refactor can bring that does to 2 requests. 

The breakdown is listed below
```
Number of DDB requests (Each * is a DDB request)
    * Read Item (dynamoDbDataService)
    (dynamoDbBundle Service)
       * Pull item id/vid
       * Lock Item
       * Stage Item
       * Unlock Item
```

carvantes (approved):

PR #57, opened 2021-02-23
feat: add support for Update as Create

### LINKED ISSUE #55: [updateCreate] Support update operation creating initial version of resource
Hi AWS team,
If we try to create a new resource via an update operation (as per https://www.hl7.org/fhir/http.html#update), the current code doesn't support this.
Previously we were doing this via a create operation but the code changes on Feb 5th (4f433d2eacdd81c25bdc6e5a2d5e9ea755a33204) now prevent us from doing this!

I'd have though it'd be relatively easy to trap the resource not found case in
https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/eefb657b0639c10910ada404f91d6418926b339c/src/dataServices/dynamoDbDataService.ts#L149
and then invoke the create operation?
Many thanks,
Allan

AllanHodkinson:
Issue #, if available:
https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/issues/55

Description of changes:
Refactored createResource & updateResource functions to support update create (https://www.hl7.org/fhir/http.html#upsert).
This is can be enabled when the Persistence component is created.


src/dataServices/dynamoDbDataService.ts
```diff
@@ -97,13 +99,16 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
         const vid = 1;
         let item = resource;
         item.resourceType = resourceType;
         item.meta = generateMeta(vid.toString());
 
-        const params = DynamoDbParamBuilder.buildPutAvailableItemParam(item, uuidv4(), vid);
+        const params = DynamoDbParamBuilder.buildPutAvailableItemParam(item, resourceId, vid);
         try {
             await this.dynamoDb.putItem(params).promise();
         } catch (e) {
```
rsmayda, line 120: on line 117 Can you re-word the error message to be more accurate; perhaps: 

"Cannot create resource since the id matched an existing id"

src/dataServices/dynamoDbDataService.ts
```diff
@@ -97,13 +99,16 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
 
     async createResource(request: CreateResourceRequest) {
         const { resourceType, resource } = request;
+        return this.createResourceWithId(resourceType, resource, uuidv4());
+    }
 
+    private async createResourceWithId(resourceType: string, resource: any, resourceId: string) {
```
rsmayda, line 106: Could we add a regex validation for resourceId? 

I believe this regex will do the job: `^[a-zA-Z0-9\-\.]{1,64}$`

[FHIR spec](https://www.hl7.org/fhir/resource.html#id): "Ids can be up to 64 characters long, and contain any combination of upper and lowercase ASCII letters, numerals, "-" and ".""

AllanHodkinson: Yes good idea.  
We could end up with a case where resources that are `created` have a different id format to those that are `updateCreated`, do we care about this?

rsmayda: From a technical stand-point I don't see any issues with a Dynamo key (id) being different or ES param (id) being different.

One issue that may come up is create a 'hot' shard meaning if all the keys started to look the same that shard of Dynamo may suffer in performance. I don't think a regex would solve this and I think anything but random can. So with us allowing `updateCreate` there will be some additional risk if users are malicious in crafting their ids

src/dataServices/dynamoDbDataService.ts
```diff
@@ -103,6 +103,11 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
     }
 
     private async createResourceWithId(resourceType: string, resource: any, resourceId: string) {
+        const regex = new RegExp('^[a-zA-Z0-9-.]{1,64}$');
+        if (!regex.test(resourceId)) {
+            throw new Error(`Resource creation failed, id ${resourceId} is not valid`);
```
rsmayda: In this case -- it would be more accurate to throw a `InvalidResourceError`; since this will only fail if folks are using `updateCreate`: https://github.com/awslabs/fhir-works-on-aws-interface/blob/mainline/src/errors/InvalidResourceError.ts

FYI this will result in a 400 bad request being returned: https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/router/routes/errorHandling.ts#L31

src/dataServices/dynamoDbDataService.test.ts
```diff
@@ -338,6 +338,32 @@ describe('UPDATE', () => {
         expect(serviceResponse.message).toEqual('Resource created');
         expect(serviceResponse.resource).toStrictEqual(expectedResource);
     });
+
+    test('ERROR: Id supplied for Update as Create is not valid', async () => {
+        // BUILD
+        const id = 'uuid:$deadbeef';
```
rsmayda, line 345: Got a good laugh here :)

src/dataServices/dynamoDbDataService.ts
AllanHodkinson: Agree.  An application could use solely `updateCreate` operations to create resources with a different style of id.  If these are not random and performance is affected then that is down to poor application design rather than a server implementation issue.

src/dataServices/dynamoDbDataService.test.ts
AllanHodkinson: That's a nod to my roots as an embedded developer!

rsmayda (approved):
Thanks for the contribution!

PR #61, opened 2021-04-06
fix: Only add customUserAgent when code is running on AWS, not when code is running locally

nguyen102:
fix: Only add `customUserAgent` when code is running on AWS, not when code is running locally


rsmayda (approved):
Need to make a change to ES/Cognito too?

nguyen102:
> Need to make a change to ES/Cognito too?

We make calls to ES using the ES Client, and not the AWS SDK, so no changes is needed there.

For Cognito, I believe we use a custom authorizer on the API GW level that passes the parsed token to the FHIR Server.

Issue #18, opened 2020-10-01
Detect all invalid resources instead of failing in Elastic Search.


arthuston:
Invalid resources are being created in DynamoDB and then failing in Elastic Search:

For example, money types like "net": 12345 are allowed when "net": { "value": 12345, "currency": "USD"} are expected. This stores the object in DynamoDB but fails during elastic search.

This occurs despite the schema validator specifying that Money is a dictionary.

AC:

Entering an invalid object should fail.

rsmayda:
Did some digging and found this: When a new object is added to an ES index all of the params of that object are dynamically mapped to an ES data type. Since `net` comes in as a string first, all future `net` must be a string or you will get this behavior. This failure causes the DDB & ES to become out of sync.

rsmayda:
As noted in https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/issues/46 "Failed records are not moved to DLQ and no logs are available. We need to add logs/DLQ to find out these type of bugs."

PrimKruskal:
@rsmayda  Is development happening for this bug ?  We have quite few scenarios not working due to data conversion issues.    Is there plans to introduce mapping templates rather than ES to determine format dynamically ? Do you require support to fix this functionality quickly ?

rsmayda:
We will start working on fixing this bug in early April. We plan to address the identification of this issue by putting in a DLQ and Cloudwatch Alarms. As for the static mapping, we do not currently have this on our roadmap, but would love to discuss more in our next stand-up.

rsmayda:
We are now logging failures into a DLQ and improved logging to identify the problematic resources

PR #68, opened 2021-04-22
fix: improve error logging for ddbToEs sync

rsmayda:
Issue #, if available: #18, https://github.com/awslabs/fhir-works-on-aws-search-es/issues/21

The goal with this PR is:
1. throw an error if the lambda can't process (causing retries/failures)
2. if errors do occur put in the logs:
    1. the resources that were 'bad'
    2. all the resources in this batch that may not have been processed

### DIFF
--- a/src/ddbToEs/index.ts
+++ b/src/ddbToEs/index.ts
@@ -25,7 +25,6 @@ export async function handleDdbToEsEvent(event: any) {
             const image = AWS.DynamoDB.Converter.unmarshall(ddbJsonImage);
             // Don't index binary files
             if (ddbToEsHelper.isBinaryResource(image)) {
-                console.log('This is a Binary resource. These are not searchable');
                 // eslint-disable-next-line no-continue
                 continue;
             }
@@ -47,6 +46,20 @@ export async function handleDdbToEsEvent(event: any) {
 
         await ddbToEsHelper.logAndExecutePromises(promiseParamAndIds);
     } catch (e) {
-        console.log('Failed to update ES records', e);
+        console.error(
+            'Synchonization failed! The resources that could be effected are: ',
+            event.Records.map(
+                (record: {
+                    eventName: string;
+                    dynamodb: { OldImage: AWS.DynamoDB.AttributeMap; NewImage: AWS.DynamoDB.AttributeMap };
+                }) => {
+                    const image = record.eventName === REMOVE ? record.dynamodb.OldImage : record.dynamodb.NewImage;
+                    return `{id: ${image.id.S}, vid: ${image.vid.N}}`;
+                },
+            ),
+        );
+
+        console.error('Failed to update ES records', e);
+        throw e;
     }
 }
--- a/src/ddbToEs/index.ts
+++ b/src/ddbToEs/index.ts
@@ -47,7 +47,7 @@ export async function handleDdbToEsEvent(event: any) {
         await ddbToEsHelper.logAndExecutePromises(promiseParamAndIds);
     } catch (e) {
         console.error(
-            'Synchonization failed! The resources that could be effected are: ',
+            'Synchronization failed! The resources that could be effected are: ',
             event.Records.map(
                 (record: {
                     eventName: string;
--- a/src/AWS.ts
+++ b/src/AWS.ts
@@ -12,13 +12,13 @@ const AWSWithXray = AWSXRay.captureAWS(AWS);
 const { IS_OFFLINE } = process.env;
 
 if (IS_OFFLINE === 'true') {
-    AWS.config.update({
+    AWSWithXray.config.update({
         region: process.env.AWS_REGION || 'us-west-2',
         accessKeyId: process.env.ACCESS_KEY,
         secretAccessKey: process.env.SECRET_KEY,
     });
 } else {
-    AWS.config.update({
+    AWSWithXray.config.update({
         customUserAgent: process.env.CUSTOM_USER_AGENT,
     });
 }
--- a/src/ddbToEs/ddbToEsHelper.ts
+++ b/src/ddbToEs/ddbToEsHelper.ts
@@ -72,7 +72,8 @@ export default class DdbToEsHelper {
                 await this.ElasticSearch.indices.create(params);
             }
         } catch (error) {
-            console.log('Failed to check if index exist or create index', error);
+            console.error(`Failed to check if index: ${indexName} exist or create index`);
+            throw error;
         }
     }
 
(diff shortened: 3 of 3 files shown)


src/ddbToEs/index.ts
```diff
@@ -47,6 +46,20 @@ export async function handleDdbToEsEvent(event: any) {
 
         await ddbToEsHelper.logAndExecutePromises(promiseParamAndIds);
     } catch (e) {
-        console.log('Failed to update ES records', e);
+        console.error(
+            'Synchonization failed! The resources that could be effected are: ',
```
carvantes: If I'm reading this correctly, this logs all records in the current batch even if only one failed. Can we isolate the error and just log the records that failed?. 
It'd be cumbersome to try to find the single record that failed among up to 15 records.

src/ddbToEs/index.ts
```diff
@@ -47,6 +46,20 @@ export async function handleDdbToEsEvent(event: any) {
...
+                    dynamodb: { OldImage: AWS.DynamoDB.AttributeMap; NewImage: AWS.DynamoDB.AttributeMap };
+                }) => {
+                    const image = record.eventName === REMOVE ? record.dynamodb.OldImage : record.dynamodb.NewImage;
+                    return `{id: ${image.id.S}, vid: ${image.vid.N}}`;
+                },
+            ),
+        );
+
+        console.error('Failed to update ES records', e);
+        throw e;
```
carvantes, line 63: I think that this will result in some messages that were processed successfully being sent to the dlq, since a single message makes the batch fail and retrying the same batch will continue to fail.

It is common to use a batch size of 1 to workaround this issue. An alternative is to enable `BisectBatchOnFunctionError` although I haven't used that setting before an I'm not sure about how it interacts with `MaximumRetryAttempts` 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html#cfn-lambda-eventsourcemapping-bisectbatchonfunctionerror

src/ddbToEs/index.ts
rsmayda: Yep that is right this log says all of these records were in the batch and that batch failed.

this log identifies the problematic ones: https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/e0b7f10bc35deda298375bcbae04feeab2f54157/src/ddbToEs/ddbToEsHelper.ts#L183-L186

rsmayda: The benefit of logging both is 1 - you see the problem ones and 2 - you know the other records that could be affected

src/ddbToEs/index.ts
rsmayda: Yep that is exactly the case some messages will succeed but the batch will fail if a single message fails.

I worry batch size of 1 may slow down our our sync too much. I looked into `BisectBatch` but as you mentioned I was not sure how it works with `MaxRetry` and I wasn't able to find documentation around it either. I suspect that it will Bisect at most MaxRetry times.

These writes are mostly idempotent, but there could be a use-case of a resource's availability switches due to this. ie 1) "AVAILIABLE" write fails and goes to DLQ, 2) "DELETE" write passes, 3) DLQ redrive changes the ES doc from DELTED -> AVAILIVABLE 

A thing to note this DLQ redrive is a manual process and in reality I suspect that this operation would need a runbook laying out when to 'redrive' the DLQ and when not to

nguyen102:
If we're making the change to throw an error, we should set the maxRetry to 1.
https://github.com/awslabs/fhir-works-on-aws-deployment/blob/a03bdc6b27fe5eca8708d415f98cd60e30b1f897/serverless.yaml#L119

Otherwise, if we fail to update ES because we have too many open sockets, then the retries will make it even worse.

Also, out of curiosity, how will this link up with pushing items to the DLQ. Is there a mechanism that will automatically push the record of failed requests to the DLQ?

rsmayda:
In reply to @nguyen102:
> maxRetry to 1.

In the case of ES throttling we would want to retry right? Perhaps it is better for the lambda itself to do retries? 

>  how will this link up with pushing items to the DLQ. Is there a mechanism that will automatically push the record of failed requests to the DLQ?

The EventStream tries `maxRetry` number of times if the lambda fails past that - a message is dropped into the DLQ containing the EventStream batch information

src/ddbToEs/index.ts
carvantes: sounds good, I didn't notice the exact erro was logged elsewhere

nit: Typo in `Synchonization`

nguyen102:
> In reply to @rsmayda :
> 
> > maxRetry to 1.
> 
> In the case of ES throttling we would want to retry right? Perhaps it is better for the lambda itself to do retries?
Yep, agreed. We should have lambda, itself, retry, instead of having the stream invoke the lambda again.

> The EventStream tries `maxRetry` number of times if the lambda fails past that - a message is dropped into the DLQ containing the EventStream batch information

Ah, got it. I assume the change for the DLQ will be in the `deployment` package then?

rsmayda:
> @nguyen102: Ah, got it. I assume the change for the DLQ will be in the `deployment` package then?

Yep: https://github.com/awslabs/fhir-works-on-aws-deployment/pull/295

src/ddbToEs/index.ts
rsmayda:
```suggestion
            'Synchronization failed! The resources that could be effected are: ',
```

rsmayda:
> @nguyen102 Yep, agreed. We should have lambda, itself, retry, instead of having the stream invoke the lambda again.

I think this would be a separate PR?  This PR is focusing on improving logging -- happy to discuss otherwise too

nguyen102:
> > @nguyen102 Yep, agreed. We should have lambda, itself, retry, instead of having the stream invoke the lambda again.
> 
> I think this would be a separate PR? This PR is focusing on improving logging -- happy to discuss otherwise too

Yep, agreed. I'm ok with it being in a separate PR. Until then should we change the `maxRetry` in the deployment package to have a value of 1. From my experience, ES error usually occurs from Lambda opening trying to open too many socket connections. Having the retries will make it worse.

src/ddbToEs/index.ts
carvantes: I think that guaranteeing that only the failed messages go to the DLQ is a very desirable property of the system. Otherwise ops become harder for customers for no good reason (why are there so many DLQ messages? How come only 6% of them actually failed? How can I know which of them actually failed?)

Another desirable property is handling out of order messages. Our current implementation does not do that(not the same as idempotency). It could be achieved by updating ES only if the vid of the incoming message is higher than the vid of the document in ES. This would make it safe to redrive DLQ messages. I think we can tackle this later as a separate issue.

IMO sending only the failed messages to the DLQ should be done now (can still be a different PR). I agree that `BisectBatch` has scarce documentation, but is worth testing it out. Maybe `MaxRetry=4` and `BisectBatch=true` with our `BatchSize=15` will effectively isolate the error to a single record.

The cheap alternative is `MaxRetry=1`.
> I worry batch size of 1 may slow down our our sync too much

My intuition tells me the same, but we need data in order to discard that approach

carvantes:
>We should have lambda, itself, retry, instead of having the stream invoke the lambda again.

IMO we have a correct setup right now. Lambda code has zero retries, and the `EventSourceMapping` has `MaximumRetryAttempts=3`. Are you suggesting to have Lambda code do retries? If so,  What's the benefit we are pursuing in doing so? It'd make the e2e retry behavior more complex.

---

> Having the retries will make it worse.

Why?

It's very likely that https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/69 is the cause of the `EMFILE` errors (creating lots and lots of ES client is a bad idea). Plus, retries don't increase the number of concurrent requests to ES. Streams MUST be processed in order, so retries on a batch will block the processing of all the other incoming messages until max retries are reached.

carvantes (approved):
Approving since this is a step in the right direction in terms of error logging.

It may be a good idea to sync up offline about the retry & DLQ setup since there are many comments about it.

nguyen102:
> Streams MUST be processed in order, so retries on a batch will block the processing of all the other incoming messages until max retries are reached.

Oh I see, assuming https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/69 is the solution to the `EMFILE` errors, that would alleviate my concern for the retries. Good point about the retry not causing `EMFILE` errors to get worse. Without the retries we would still have `EMFILE` errors as the lamda process new batches and once again encounter the error. Although the retry does cause the `EMFILE` error to occur for 3 times longer if we do get to this state.

I'm ok, with approving this PR and merging in https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/69. We can gather more data after this and figure out best step forward.

nguyen102 (approved):

PR #69, opened 2021-04-22
feat!: expect elastic search instance to be passed in

rsmayda:
BREAKING CHANGE

Description of changes:
- Following the AWS best practice:
> AWS service clients should be instantiated in the initialization code, not in the handler. This allows AWS Lambda to reuse existing connections, for the duration of the container's lifetime.
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.Lambda.BestPracticesWithDynamoDB.html

### DIFF
--- a/src/ddbToEs/ddbToEsHelper.ts
+++ b/src/ddbToEs/ddbToEsHelper.ts
@@ -4,39 +4,18 @@
  */
 
 import { Client } from '@elastic/elasticsearch';
-// @ts-ignore
-import { AmazonConnection, AmazonTransport } from 'aws-elasticsearch-connector';
 import allSettled from 'promise.allsettled';
-import AWS from '../AWS';
 import PromiseParamAndId, { PromiseType } from './promiseParamAndId';
 import { DOCUMENT_STATUS_FIELD } from '../dataServices/dynamoDbUtil';
 import DOCUMENT_STATUS from '../dataServices/documentStatus';
 
 const BINARY_RESOURCE = 'binary';
 
-const { IS_OFFLINE, ELASTICSEARCH_DOMAIN_ENDPOINT } = process.env;
-
 export default class DdbToEsHelper {
     private ElasticSearch: Client;
 
-    constructor() {
-        let ES_DOMAIN_ENDPOINT = ELASTICSEARCH_DOMAIN_ENDPOINT || 'https://fake-es-endpoint.com';
-        if (IS_OFFLINE === 'true') {
-            const { ACCESS_KEY, SECRET_KEY, AWS_REGION, OFFLINE_ELASTICSEARCH_DOMAIN_ENDPOINT } = process.env;
-
-            AWS.config.update({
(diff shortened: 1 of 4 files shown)


src/ddbToEs/ddbToEsHelper.ts
```diff
@@ -4,39 +4,18 @@
...
-            });
-            ES_DOMAIN_ENDPOINT = OFFLINE_ELASTICSEARCH_DOMAIN_ENDPOINT || 'https://fake-es-endpoint.com';
-        }
-
-        this.ElasticSearch = new Client({
-            node: ES_DOMAIN_ENDPOINT,
-            Connection: AmazonConnection,
-            Transport: AmazonTransport,
-        });
+    constructor(ElasticSearch: Client) {
```
rsmayda, line 20: Do we want to do this across the package? i.e. Deployment package initializes the AWS services and passes it down?

src/ddbToEs/index.ts
```diff
@@ -13,9 +14,9 @@ const REMOVE = 'REMOVE';
 // This lambda picks up changes from DDB by way of DDB stream, and sends those changes to ElasticSearch Service for indexing.
 // This allows the FHIR API Server to query ElasticSearch service for search requests
 
-export async function handleDdbToEsEvent(event: any) {
-    const ddbToEsHelper = new DdbToEsHelper();
+export async function handleDdbToEsEvent(event: any, ElasticSearch: Client) {
```
nguyen102, line 19: Nitpick for naming: `ElasticSearch => esClient`

carvantes, line 19: I think that this function shouldn't take the client as a param. It's awkward to pass the client to every single invocation. 

It can be changed to a class that takes the `esClient` as a param and keeps a reference to it. Then we can keep the original fn signature`handleDdbToEsEvent(event: any) => {...}`

That's similar to how other fwoa components work, with the difference that they don't take aws clients as params but they still init clients only once in constructors.

src/ddbToEs/ddbToEsHelper.ts
carvantes: I'd argue that we should not make this change as part of this PR. It's a bit too extreme and it's not strictly needed to fix the issue.

we could do something like:
```js
class ourNewClass {
  private ElasticSearch: Client;
  constructor(){
    this.elasticsearch = new Client()
  }
  handleDdbToEsEvent(event: any) {
    ...
  }
}
``` 
to bring it in line with all the other fwoa packages(they tend to init aws clients in constructors)

I think that later we can evaluate the option of having fwoa components accept aws clients as optional constructor params. IMO optional is better since customers often only want to pass down clients if they have a very specific client config in mind (i.e. very specific retry policies, overriding endpoints, etc) otherwise it's convenient to just use our default AWS client.

rsmayda:
Closing in favor of this implementation: https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/79

PR #73, opened 2021-05-10
feat: DynamoDB persistence support for archiving resources

Zambonilli:
Issue #, if available:
[290](https://github.com/awslabs/fhir-works-on-aws-deployment/issues/290)

Description of changes:
Added support for archiving of resources using a persistence TTL value. For dynamoDB, this is native and added via the new `ttlInSeconds` value. When resources are created or updated, a new ttlInSeconds is passed in and stored in the ttlInSeconds field. One note, the update path needs to also update all old versions so all versions of resource instance are expired at the same time and flagged for archiving.

The interface changes were merged to mainline with this [PR](https://github.com/awslabs/fhir-works-on-aws-interface/pull/64).

The routing and deployment changes to add support for archiving are in my forks and I am submitting PRs cascading down the dependency trees.
[routing](https://github.com/Zambonilli/fhir-works-on-aws-routing)
[deployment](https://github.com/Zambonilli/fhir-works-on-aws-deployment)


src/dataServices/dynamoDbUtil.ts
```diff
@@ -29,10 +30,19 @@ export class DynamoDbUtil {
         return cleanedItem;
```
nguyen102, line 32: I think we're going to need to update this method to remove `ttlInSeconds` from the DDB record. `ttlInSeconds` is system generated meta data similar to `LOCK_END_TS_FIELD` and should not be returned to the user.

While we're here, should we make a TTL_IN_SECONDS_FIELD constant so we can reuse that value across classes?

nguyen102 (commented):
Overall looks good. One small comment.

src/dataServices/dynamoDbUtil.ts
Zambonilli: Since ttlInSeconds is being configured and computed upstream in deployment and routing, I returned the value on resource instances from persistence. Then in the routing, I removed it from the output json crudely. I figured long term, there's going to be more internal backend logic data only so we'll need better adapter patterns in routing to guarantee what's going out the door is only valid FHIR response payloads. If that doesn't make sense or builds up some unwanted tech debt, let me know and I can easily update to remove ttlInSeconds in the clean method.

Agreed on the constant here and will add.

Zambonilli: Pushed the constant and bumped the interface version to 8.2.0 to pull in the ttlInSeconds.

nguyen102: Thanks for the constant update and interface version update. 

In terms of using the `ttlInSeconds` value, does the routing package need to use this value? I'm thinking this value can be handled completely by the `persistence` package. We can add `ttlInSeconds` as an optional parameter that is passed to the constructor of `DynamoDbDataService`. If `ttlInSeconds` exist then we append that value to Create and Update requests. If not then we keep the original behavior of the class.

Zambonilli: Sure, the persistence constructors could take in the Map<resourceType:string, ttlInSeconds> and move the compute logic to persistence. We'll need to rollback the interface changes, add logic to pass the map to persistence constructor, move the computation, remove the passing of ttlInSeconds to the persistence methods and remove the logic for removing ttlInSeconds from API responses.

I don't feel strongly here either way, more concerned with what is best long term for code quality. If it makes sense to move all the archive logic to persistence to help encapsulate ownership and keep the layers to a minimum than I'll get started on the updates.

Zambonilli: @nguyen102 Updated interface, persistence, routing and deployment to have the persistence layer handle ttls. I have all of the repos up on my AWS Account but ran out of time to start testing today. I will try to have testing and any bug fixes ready by Monday morning for this PR.

Zambonilli: Finished up manual testing against our AWS Account's FHIR Works instance. The ttl change to being managed in persistence looks fine.

nguyen102: @zambonilli Just to double check, what interface change was required to make to have persistence handle ttl? I was under the impression that no change was needed, but would like to double check with you in case I missed something.

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -422,6 +426,35 @@ export class DynamoDbBundleService implements Bundle {
...
+                                updateRequest.Put.Item.id.S,
+                                vid,
+                                updateRequest.Put.Item.resourceType.S,
+                                parseInt(updateRequest.Put.Item.ttlInSeconds.N, 10),
+                            );
+                        });
+                    }),
+                );
+
+                await this.dynamoDb
```
nguyen102: `transactWriteItems` have a maximum limit of 25 actions per request. 
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html

We should check how many updates there are and break it into groups of 25 if needed. We also need to keep track of which record was updated and roll them back if one of the groups of 25 failed to be updated.

nguyen102 (commented):
Almost there. One comment about updates for Batch. Also a comment from the deployment package to move the firehose lambda to this package.

src/dataServices/dynamoDbBundleService.ts
Zambonilli: Yeah, I was worried about this too. However, when checking upstream the routing is returning a 400 when more than 25 entries are requested. If we want to be more robust persistence than upstream routing, I can chunk by 25 and submit to ddb.

https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/router/bundle/bundleHandler.ts#L135

nguyen102: This is for `ttlInSecondsUpdates`, right?

What if there was 8 bundle update requests, which is under 25, but each Update request is updating a record with 4 versions. In that case we would have 32 entries in `ttlInSecondsUpdates`.

Zambonilli: Oh shoot, you're right. Sorry, in my head the update was 1:1. I'll update this to chunk to 25 per update call.

nguyen102: no worries. Let me know when you've moved the firehose lambda code to this PR.

Zambonilli: @nguyen102 pushed the fix here to chunk and ported dynamoDBToFirehose logic to persistence-ddb. The deployment PR is failing until persistence-ddb version is bumped and published to pickup the export of the function.

src/dataServices/dynamoDbBundleService.test.ts
```diff
@@ -320,9 +326,11 @@ describe('atomicallyReadWriteResources', () => {
                 startTime: new Date(),
             });
 
+            newResource.ttlInSeconds = Math.floor(Date.now() / 1000 + 60);
+
             // CHECK
-            // transactWriteItem requests is called thrice
-            expect(transactWriteItemSpy.calledThrice).toBeTruthy();
+            // transactWriteItem requests is called 4x
+            expect(transactWriteItemSpy.callCount).toEqual(4);
```
rsmayda: So I see we now have test cases with ttl Set do we have test cases without ttl set?

src/dataServices/dynamoDbBundleService.test.ts
```diff
@@ -377,8 +385,26 @@ describe('atomicallyReadWriteResources', () => {
...
 
-            // 2. change Patient record's documentStatus to be 'AVAILABLE'
+            // 2. Update ttlInSeconds for previous version
             expect(transactWriteItemSpy.getCall(2).args[0]).toStrictEqual({
+                TransactItems: [
+                    {
+                        Update: {
+                            TableName: '',
+                            Key: { id: { S: id }, vid: { N: oldVid.toString() } },
+                            UpdateExpression: 'set ttlInSeconds = :ttlInSeconds',
```
rsmayda: from a ddb perspective should we have all 'private'/'non-fhir' data elements prefixed with an underscore '_'? So `ttlInSeconds` -> `_ttlInSeconds`

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -24,7 +25,11 @@ export interface ItemRequest {
 }
 
 export default class DynamoDbBundleServiceHelper {
-    static generateStagingRequests(requests: BatchReadWriteRequest[], idToVersionId: Record<string, number>) {
+    static generateStagingRequests(
+        requests: BatchReadWriteRequest[],
+        idToVersionId: Record<string, number>,
+        ttlsInSeconds: Map<string, number>,
```
rsmayda: can this default ?
```suggestion
        ttlsInSeconds: Map<string, number> = new Map<string, number>(),
```

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -42,16 +47,27 @@ export default class DynamoDbBundleServiceHelper {
...
 
                     createRequests.push({
                         Put: {
                             TableName: RESOURCE_TABLE,
                             Item: DynamoDBConverter.marshall(Item),
                         },
                     });
+
+                    const stagingResource = _.cloneDeep(request.resource);
+                    if (_.has(Item, 'ttlInSeconds')) {
```
rsmayda: should 'ttlInSeconds' also be the constant `TTL_IN_SECONDS`?

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -73,8 +95,13 @@ export default class DynamoDbBundleServiceHelper {
                         },
                     });
 
+                    const stagingResource = _.cloneDeep(request.resource);
+                    if (_.has(Item, 'ttlInSeconds')) {
```
rsmayda: should 'ttlInSeconds' also be the constant TTL_IN_SECONDS?

src/dataServices/dynamoDbUtil.ts
```diff
@@ -29,10 +32,20 @@ export class DynamoDbUtil {
...
+        vid: number,
+        documentStatus: DOCUMENT_STATUS,
+        ttlInSeconds?: number,
+    ) {
         const item = clone(resource);
         item.id = id;
         item.vid = vid;
+        if (!_.isUndefined(ttlInSeconds)) {
+            const unixNow: number = Math.floor(Date.now() / 1000);
+            item[TTL_IN_SECONDS] = unixNow + ttlInSeconds;
```
rsmayda: interesting so this is more of expiration time than TTL?

rsmayda (commented):
First pass will look again

src/dataServices/dynamoDbUtil.ts
Zambonilli: The dynamo DB documentation defines it as a TTL so I tried to follow suit. When people pass the map to the constructor it is actually a ttl but dynamodb expects an expiry date.

src/dataServices/dynamoDbBundleServiceHelper.ts
Zambonilli: yep, ty. pushed an update

src/dataServices/dynamoDbBundleServiceHelper.ts
Zambonilli: yep, ty. pushed an update

src/dataServices/dynamoDbBundleServiceHelper.ts
Zambonilli: Yeah, I didn't because I defaulted in the persistence and bundle services constructors. However, it did feel weird updating all the helper tests. I'll update to optional and remove from all tests that don't assert the archive logic.

Zambonilli: pushed an update to make this optional and default. Also, rolled back the tests that were passing in an empty map.

src/dataServices/dynamoDbBundleService.test.ts
Zambonilli: We could just change the constant here in persistence and the resource table's ttl configuration. I'd prefer to not rename across all the code though because I've already done that 1x already. :-)

src/dataServices/dynamoDbBundleService.test.ts
Zambonilli: We don't, mainly because the test setup code was pretty rough. I'll update the `runUpdateTest` method to take in another bool to be able to toggle whether to include ttlInSeconds.

Long term, I was going to open another issue and do a PR to get a builder pattern going for the test cases. This would allow us to build permutations of test setup a lot easier and with a fluid api. http://natpryce.com/articles/000714.html

Zambonilli: pushed 2 new test cases for bundle update to test when there is not a ttlInSeconds set

rsmayda: Thank you!

rsmayda:
Hey @Zambonilli, This is our first big independent feature outside of our team and our known use-cases. So after speaking within our team and to some of our customers we want to do a Request for Comment (RFC) cycle. A RFC is when we ask our community:

1. Who would use this TTL change?
2. Should this be in the core repo enabled by default; or disabled by default; or in a different repo?
3. What is the business benefit (or provide a real world example) of the TTL change?

We (FWoA) will create an issue in the deployment package trying to answer these questions (by the end of this week). Once posted (or before) we can review the answers with you to ensure the accuracy. From there we will leave the issue open for a few weeks and decide our path forward from there.

We appreciate your contributions and we don't want this to block you. We would be happy to set some time up to discuss this process and talk about next steps. Thank you for being an active member of the community!

rsmayda (changes requested):
Pending RFC review

src/dataServices/dynamoDbBundleService.test.ts
Zambonilli: Updated the field name to _ttlInSeconds to match _resources convention. I should have caught this earlier. Also, so someone else doesn't hit this problem, dynamoDB's update expression doesn't allow field names with `_` characters. You have to use the ExpressionAttributeNames to map the field name per https://github.com/aws/aws-sdk-js/issues/1691

Zambonilli:
@rsmayda No problem, I totally understand and please do whatever is best for the long term. We've been talking with our AWS solution architects and some members of the FWoA team already. I will send this link to them to see if they can setup a meeting with you guys directly on how best we can help advance FWoA.

To answer your RFC questions above directly.
1.  We're a medtech company.
2. Current implementation is in the core repos and optional by default.
3. We're using the archiving functionality to support our privacy and regulatory requirement of keeping audit events online for only 6 months in the EU. This is for GDPR and I believe HDS.

Zambonilli:
Closing per comments in https://github.com/awslabs/fhir-works-on-aws-deployment/issues/340 about a plan to make archiving an optional and sandboxed from normal transaction processing flow feature set.

PR #79, opened 2021-05-19
fix: initialize ES client only once

Bingjiling:
Issue #, if available:

Description of changes:

* The root cause of this issue is the connections used for ES being held up in cache after Lambda function done running. And the sockets used for the connection takes up the file descriptors allocated to the Lambda function. Lambda function has a hard limit for number of file descriptors it can use: https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html 
* The solution here is to initialize the ES client only once for the container. This way cached connections can be reused, and only a constant number of file descriptors(~450) are needed. 
* The change is load tested and esToDDB Lambda can sync the data consistently when hitting FHIR server at 45 TPS

### DIFF
--- a/src/ddbToEs/index.ts
+++ b/src/ddbToEs/index.ts
@@ -10,13 +10,13 @@ import getComponentLogger from '../loggerBuilder';
 
 const REMOVE = 'REMOVE';
 const logger = getComponentLogger();
+const ddbToEsHelper = new DdbToEsHelper();
 
 // This is a separate lambda function from the main FHIR API server lambda.
 // This lambda picks up changes from DDB by way of DDB stream, and sends those changes to ElasticSearch Service for indexing.
 // This allows the FHIR API Server to query ElasticSearch service for search requests
 
 export async function handleDdbToEsEvent(event: any) {
-    const ddbToEsHelper = new DdbToEsHelper();
     try {
         const promiseParamAndIds: PromiseParamAndId[] = [];
         for (let i = 0; i < event.Records.length; i += 1) {


yarn.lock
```diff
@@ -2161,10 +2161,8 @@ fecha@^4.2.0:
   resolved "https://registry.yarnpkg.com/fecha/-/fecha-4.2.1.tgz#0a83ad8f86ef62a091e22bb5a039cd03d23eecce"
   integrity sha512-MMMQ0ludy/nBs1/o0zVOiKTpG7qMbonKUzjJgQFEuvq6INZ1OraKPRAWkBq5vlKLOUMpmNYG1JoN3oDPUQ9m3Q==
 
-fhir-works-on-aws-interface@^9.0.0:
```
rsmayda: revert this

src/ddbToEs/index.ts
```diff
@@ -63,5 +63,7 @@ export async function handleDdbToEsEvent(event: any) {
 
         logger.error('Failed to update ES records', e);
         throw e;
+    } finally {
+        await ddbToEsHelper.closeConnection();
```
carvantes: This does fix the issue but I don't think it is the best solution.

What happened with https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/69#discussion_r618701347? Having exactly one instance of the ES client would fix this issue, and give the best performance since you'd still get the benefit of connection reuse.

As it stands now, the lambda function is still initializing an ES client for every invocation, and that's a bad practice.

carvantes (changes requested):

src/ddbToEs/index.ts
Bingjiling: I tried using codes from #69 but it was throwing errors so I switched back to mainline. I can try it out again. But with this fix, the ddbToEs Lambda is no longer the bottleneck.

nguyen102:
Nice find Yanyu. Btw, you can get ride of the 429 error limit by changing the rate limit here.
https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/serverless.yaml#L62

Although, I'm surprised that we're getting 429 errors when we're only at 45 tps.

src/ddbToEs/index.ts
Bingjiling: Just finished testing changes in #69, it does fix the issue, but I decided to stick to this fix as it does not bring breaking changes.

carvantes (approved):
LGTM, just make sure to update the PR title/description to reflect the latest changes.

PR #80, opened 2021-05-25
feat: support update as create in bundle service

justinusmenzel:
Description of changes:
In order to lock resources for operations other than 'create' the latest version ID needs to be looked up first.  
If the resource isn't found a ResourceNotFoundError is thrown. 
With this change only ResourceNotFoundErrors for operations other than 'update' are being counted and result in a failed transaction.
By allowing 'update' operation for resources that don't exist in dynamo yet using update as create is possible.

### DIFF
--- a/src/dataServices/dynamoDbBundleService.test.ts
+++ b/src/dataServices/dynamoDbBundleService.test.ts
@@ -8,7 +8,7 @@ import * as AWSMock from 'aws-sdk-mock';
 
 import { QueryInput, TransactWriteItemsInput } from 'aws-sdk/clients/dynamodb';
 import AWS from 'aws-sdk';
-import { BundleResponse, BatchReadWriteRequest } from 'fhir-works-on-aws-interface';
+import { BundleResponse, BatchReadWriteRequest, TypeOperation } from 'fhir-works-on-aws-interface';
 import { DynamoDbBundleService } from './dynamoDbBundleService';
 import { DynamoDBConverter } from './dynamoDb';
 import { timeFromEpochInMsRegExp, utcTimeRegExp, uuidRegExp } from '../../testUtilities/regExpressions';
@@ -449,5 +449,56 @@ describe('atomicallyReadWriteResources', () => {
         test('UPDATING a resource with references', async () => {
             await runUpdateTest(true);
         });
+    });
+
+    describe('Update as Create Cases', () => {
+        const runTest = async (supportUpdateCreate: boolean, operation: TypeOperation, isLockSuccessful: boolean) => {
+            // READ items (Failure)
+            AWSMock.mock('DynamoDB', 'query', (params: QueryInput, callback: Function) => {
+                callback(null, { Items: [] });
+            });
+
+            const dynamoDb = new AWS.DynamoDB();
+            const bundleService = new DynamoDbBundleService(dynamoDb, supportUpdateCreate);
+
+            const updateRequest: BatchReadWriteRequest = {
+                operation,
+                resourceType: 'Patient',
+                id,
+                resource: 'Patient/bce8411e-c15e-448c-95dd-69155a837405',
+            };
+            // @ts-ignore
+            const actualResponse = await bundleService.lockItems([updateRequest]);
+            if (isLockSuccessful) {
+                expect(actualResponse).toStrictEqual({
+                    lockedItems: [],
+                    successfulLock: true,
+                });
+            } else {
+                expect(actualResponse).toStrictEqual({
+                    errorMessage: 'Failed to find resources: Patient/bce8411e-c15e-448c-95dd-69155a837405',
+                    errorType: 'USER_ERROR',
+                    lockedItems: [],
+                    successfulLock: false,
+                });
+            }
+        };
+
+        const testCases = [
+            // ['supportUpdateCreate', 'operation', 'isLockSuccessful'],
+            [true, 'create', true],
+            [true, 'update', true],
+            [true, 'read', false],
+            [false, 'create', true],
+            [false, 'update', false],
+            [false, 'read', false],
+        ];
+
+        // eslint-disable-next-line no-restricted-syntax
+        for (const [supportUpdateCreate, operation, isLockSuccessful] of testCases) {
+            test('lock update ', async () => {
+                await runTest(supportUpdateCreate as boolean, operation as TypeOperation, isLockSuccessful as boolean);
+            });
+        }
     });
 });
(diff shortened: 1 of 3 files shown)


src/dataServices/dynamoDbBundleService.ts
```diff
@@ -190,7 +190,8 @@ export class DynamoDbBundleService implements Bundle {
         const idItemsFailedToRead: string[] = [];
         for (let i = 0; i < itemResponses.length; i += 1) {
             const itemResponse = itemResponses[i];
-            if (itemResponse instanceof ResourceNotFoundError) {
+            // allow for update as create scenario
+            if (itemResponse instanceof ResourceNotFoundError && itemsToLock[i].operation !== 'update') {
```
carvantes: I think that this works, but it will inadvertently enable update as crate for all customers using bundles/transactions.

`DynamoDbDataService` has a parameter `supportUpdateCreate` that enables update as create for the regular update operation but doesn't address bundles/transactions. I think that we can pass down that same parameter to `DynamoDbBundleService` and use it to conditionally apply your changes.

https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/e913c711c842d922a9aa1902b6705d240af6ad68/src/dataServices/dynamoDbDataService.ts#L71

justinusmenzel: I just added the `supportUpdateCreate` flag to dynamoDbBundleService.ts. Creating a test seems not so trivial. I'll give it a shot.

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -190,7 +193,11 @@ export class DynamoDbBundleService implements Bundle {
         const idItemsFailedToRead: string[] = [];
         for (let i = 0; i < itemResponses.length; i += 1) {
             const itemResponse = itemResponses[i];
-            if (itemResponse instanceof ResourceNotFoundError) {
+            // allow for update as create scenario
+            if (
+                itemResponse instanceof ResourceNotFoundError &&
+                !(itemsToLock[i].operation === 'update' && this.updateCreateSupported)
```
rsmayda, line 199: nit: I would flip the `if` block to do the Boolean check first 
```suggestion
                this.updateCreateSupported && itemResponse instanceof ResourceNotFoundError &&
                !(itemsToLock[i].operation === 'update')
```

justinusmenzel: The two expressions don't seem to evaluate the same. Now if `this.updateCreateSupported` is false the whole expression evaluates to false no matter what the operation is or if there's an error or not. Am I missing something?

rsmayda: oh yep you are right there are parenthesis, (missed that)

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -190,7 +193,11 @@ export class DynamoDbBundleService implements Bundle {
...
+            ) {
                 idItemsFailedToRead.push(`${itemsToLock[i].resourceType}/${itemsToLock[i].id}`);
```
rsmayda, line 201: So allowing `updateAsCreate` what is the versionId created in this scenario? will it take whatever is in the resource or will the server generate it?

justinusmenzel: I don't think there's a way for the resource in a bundle to provide a vid. If it had one it would just be ignored.
If the vid can't be found in the `idToVersionId` map (which is populated with the data coming back from lockItems) it's assumed to be 1: https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/56c16a44692b892a042baa92b53ccc4e661ae146/src/dataServices/dynamoDbBundleServiceHelper.ts#L66

src/dataServices/dynamoDbBundleService.test.ts
```diff
@@ -450,4 +450,55 @@ describe('atomicallyReadWriteResources', () => {
             await runUpdateTest(true);
         });
     });
+
+    describe('Update as Create Cases', () => {
```
rsmayda, line 454: These look great could you run: `yarn run test --coverage` this will show any uncovered lines. You can just share the `src/dataServices/dynamoDbBundleService.ts` row if you want

src/dataServices/dynamoDbBundleService.test.ts
```diff
@@ -450,4 +450,55 @@ describe('atomicallyReadWriteResources', () => {
...
+        const runTest = async (supportUpdateCreate: boolean, operation: TypeOperation, isLockSuccessful: boolean) => {
+            // READ items (Failure)
+            AWSMock.mock('DynamoDB', 'query', (params: QueryInput, callback: Function) => {
+                callback(null, { Items: [] });
+            });
+
+            const dynamoDb = new AWS.DynamoDB();
+            const bundleService = new DynamoDbBundleService(dynamoDb, supportUpdateCreate);
+
+            const updateRequest: BatchReadWriteRequest = {
```
rsmayda: This could be a non-update request right?

```suggestion
            const batchRequest: BatchReadWriteRequest = {

```

src/dataServices/dynamoDbBundleService.test.ts
```diff
@@ -450,4 +450,55 @@ describe('atomicallyReadWriteResources', () => {
...
+            // @ts-ignore
+            const actualResponse = await bundleService.lockItems([updateRequest]);
+            if (isLockSuccessful) {
+                expect(actualResponse).toStrictEqual({
+                    lockedItems: [],
+                    successfulLock: true,
+                });
+            } else {
+                expect(actualResponse).toStrictEqual({
+                    errorMessage: 'Failed to find resources: Patient/bce8411e-c15e-448c-95dd-69155a837405',
```
rsmayda:
```suggestion
                    errorMessage: `Failed to find resources: Patient/${id}`,

```

src/dataServices/dynamoDbBundleService.test.ts
```diff
@@ -450,4 +450,55 @@ describe('atomicallyReadWriteResources', () => {
...
+            });
+
+            const dynamoDb = new AWS.DynamoDB();
+            const bundleService = new DynamoDbBundleService(dynamoDb, supportUpdateCreate);
+
+            const updateRequest: BatchReadWriteRequest = {
+                operation,
+                resourceType: 'Patient',
+                id,
+                resource: 'Patient/bce8411e-c15e-448c-95dd-69155a837405',
```
rsmayda:
```suggestion
                resource: `Patient/${id}`,

```

src/dataServices/dynamoDbBundleService.test.ts
justinusmenzel: There are some uncovered lines, overall pretty good coverage:
`dynamoDbBundleService.ts           |   85.53 |    75.38 |   88.89 |   85.81 | 54,60,76-80,109-113,162-164,188,323-328,354-355,373-375,428-429`

rsmayda: Great your changed lines are covered :)

rsmayda (approved):
Looks good to me! @carvantes for final review and merge

carvantes (approved):

PR #83, opened 2021-06-08
feat: add alias for new and existing index

Bingjiling:
Issue #, if available:

Description of changes:

* Add alias for new and existing index on update/create
* Use alias for update/create
* Tested for 
  * Alias is created for existing index upon update/create
  * New index is created with alias 
  * Create, update and delete patient using index alias 

### DIFF
--- a/src/ddbToEs/ddbToEsHelper.test.ts
+++ b/src/ddbToEs/ddbToEsHelper.test.ts
@@ -35,7 +35,7 @@ describe('DdbToEsHelper', () => {
                 mockAddIndex,
             );
             // TEST
-            await ddbToEsHelper.createIndexIfNotExist('patient');
+            await ddbToEsHelper.createIndexAndAliasIfNotExist('patient');
             // VALIDATE
             expect(mockAddIndex).toHaveBeenCalledWith(
                 expect.objectContaining({
@@ -80,7 +80,7 @@ describe('DdbToEsHelper', () => {
                 mockAddAlias,
             );
             // TEST
-            await ddbToEsHelper.createIndexIfNotExist('patient');
+            await ddbToEsHelper.createIndexAndAliasIfNotExist('patient');
             // VALIDATE
             expect(mockAddAlias).toHaveBeenCalledWith(
                 expect.objectContaining({ path: '/patient/_alias/patient-alias' }),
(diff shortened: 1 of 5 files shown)


src/ddbToEs/ddbToEsHelper.ts
```diff
@@ -43,8 +43,10 @@ export default class DdbToEsHelper {
     }
 
     async createIndexIfNotExist(indexName: string) {
+        logger.info('entering create index function');
```
carvantes: change to debug level or remove.

src/ddbToEs/ddbToEsHelper.ts
```diff
@@ -70,12 +72,27 @@ export default class DdbToEsHelper {
                                 },
                             },
                         },
+                        [`${indexName}-alias`]: {},
                     },
                 };
                 await this.ElasticSearch.indices.create(params);
+            } else {
```
carvantes, line 79: I think that the if conditions are slightly off.
If the index does not exists, then we should create the index AND the alias. (right now the alias is only created if the index existed before this method call)

something like:
```js
if (aliasExists){
	return
}

if(!indexExists){
	createIndex()
}

createAlias()
```

src/ddbToEs/ddbToEsHelper.test.ts
```diff
@@ -0,0 +1,90 @@
+import { Client } from '@elastic/elasticsearch';
+
+import Mock from '@elastic/elasticsearch-mock';
```
carvantes, line 3: 👍

carvantes (changes requested):
Looks good so far, but I left a comment that needs to be addressed before merging.

Also, we need to make a change to write to aliases instead of indexes at: https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/0da8a2866346ad54929c0efc7fdbbb61d957c1eb/src/ddbToEs/ddbToEsHelper.ts#L127

Otherwise the aliases are meaningless.

src/ddbToEs/ddbToEsHelper.ts
Bingjiling: The create index and alias scenario is handled by line 75, where the alias is created as part of the create index call.

carvantes: My bad.

Is line 75 correct? Shouldn't it be inside a "aliases" object?
i.e.
```js
"aliases": {
    [`${indexName}-alias`]: {},
  }
```

https://www.elastic.co/guide/en/elasticsearch/reference/7.x/indices-create-index.html#create-index-aliases

Bingjiling: Good catch!

carvantes (approved):

PR #86, opened 2021-06-15
feat: tenant isolation for DDB CRUD operations

carvantes:
Use tenantId as part of the hash key for all interactions with DDB.

The main changes are in `src/dataServices/dynamoDbUtil.ts` and `src/dataServices/dynamoDbParamBuilder.ts`. everything else is mostly just passing the tenantId around.

An example DDB record with multitenancy enabled would be:
```json
{
    "id": "tenant1|id1",
    "vid": 1,
    "_id": "id1",
    "_tenantId": "tenant1",
    "resourceType": "Patient",
    "birthDate": "1990-01-01"
    ...
}
```

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -3,3 +3,4 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 export const SEPARATOR: string = '_';
+export const DDB_HASH_KEY_SEPARATOR: string = '|';
--- a/src/dataServices/dynamoDbBundleService.ts
+++ b/src/dataServices/dynamoDbBundleService.ts
@@ -39,23 +39,42 @@ export class DynamoDbBundleService implements Bundle {
 
     private maxExecutionTimeMs: number;
 
+    readonly enableMultiTenancy: boolean;
+
     private static readonly dynamoDbMaxBatchSize = 25;
 
     // Allow Mocking DDB
-    constructor(dynamoDb: DynamoDB, supportUpdateCreate: boolean = false, maxExecutionTimeMs?: number) {
+    constructor(
+        dynamoDb: DynamoDB,
+        supportUpdateCreate: boolean = false,
+        maxExecutionTimeMs?: number,
+        { enableMultiTenancy = false }: { enableMultiTenancy?: boolean } = {},
+    ) {
         this.dynamoDbHelper = new DynamoDbHelper(dynamoDb);
         this.dynamoDb = dynamoDb;
         this.updateCreateSupported = supportUpdateCreate;
         this.maxExecutionTimeMs = maxExecutionTimeMs || 26 * 1000;
+        this.enableMultiTenancy = enableMultiTenancy;
+    }
+
+    private assertValidTenancyMode(tenantId?: string) {
+        if (this.enableMultiTenancy && tenantId === undefined) {
+            throw new Error('This instance has multi-tenancy enabled, but the incoming request is missing tenantId');
+        }
+        if (!this.enableMultiTenancy && tenantId !== undefined) {
+            throw new Error('This instance has multi-tenancy disabled, but the incoming request has a tenantId');
+        }
     }
 
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     async batch(request: BatchRequest): Promise<BundleResponse> {
+        this.assertValidTenancyMode(request.tenantId);
         throw new Error('Batch operation is not supported.');
     }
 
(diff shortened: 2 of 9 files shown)


src/constants.ts
```diff
@@ -3,3 +3,4 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 export const SEPARATOR: string = '_';
+export const DDB_HASH_KEY_SEPARATOR: string = '|';
```
Bingjiling, line 6: Nice!

Bingjiling (approved):
LGTM

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -39,23 +39,42 @@ export class DynamoDbBundleService implements Bundle {
...
+
     private static readonly dynamoDbMaxBatchSize = 25;
 
     // Allow Mocking DDB
-    constructor(dynamoDb: DynamoDB, supportUpdateCreate: boolean = false, maxExecutionTimeMs?: number) {
+    constructor(
+        dynamoDb: DynamoDB,
+        supportUpdateCreate: boolean = false,
+        maxExecutionTimeMs?: number,
+        { enableMultiTenancy = false }: { enableMultiTenancy?: boolean } = {},
```
rsmayda, line 51: I am sadly confused by this syntax so we have 3 parameters (ddb, updateCreate, timeMS) and then an object to pass in and by default it is an empty object and has the structure of `{ enableMultiTenancy?: boolean }`?

Why just not pass in a Boolean and default to false?

src/dataServices/dynamoDbUtil.ts
```diff
@@ -5,13 +5,22 @@
...
 
 export const DOCUMENT_STATUS_FIELD = 'documentStatus';
 export const LOCK_END_TS_FIELD = 'lockEndTs';
 export const VID_FIELD = 'vid';
 export const REFERENCES_FIELD = '_references';
+export const TENANT_ID_FIELD = '_tenantId';
+export const INTERNAL_ID_FIELD = '_id';
+
+export const buildHashKey = (id: string, tenantId?: string): string => {
+    if (tenantId !== undefined) {
```
rsmayda: should this just be truthy? is an empty string `""` a valid tenantID?
```suggestion
    if (tenantId) {

```

src/dataServices/dynamoDbUtil.ts
```diff
@@ -45,6 +72,11 @@ export class DynamoDbUtil {
         item[DOCUMENT_STATUS_FIELD] = documentStatus;
         item[LOCK_END_TS_FIELD] = Date.now();
 
+        if (tenantId !== undefined) {
```
rsmayda: truthy check?
```suggestion
        if (tenantId) {

```

src/dataServices/dynamoDbBundleService.ts
carvantes: We have many optional parameters already, this leads to an awkward API when you only want to provide a single optional argument. e.g. `new DynamoDbBundleService(ddbInstance, undefined, 7000)` 

This is to stop us from creating more. The idea is that future new optional params are put as keys on the last param object, just like `enableMultiTenancy`.
 
Eventually it'd be nice to move `supportUpdateCreate` and `maxExecutionTimeMs` in to the last optional param object, but that's a breaking change that I don't want to do right now.

This is a fairly common pattern. e.g. see [axios api](https://github.com/axios/axios#request-method-aliases)

This article sums it up: https://codeutopia.net/blog/2016/11/24/best-practices-for-javascript-function-parameters/#:~:text=Objects%20are%20good%20for%20optionals

carvantes (commented):
Also updated to use truthy checks. (Not sure why I can't reply to the comments directly)

src/dataServices/dynamoDbUtil.ts
carvantes: updated

src/dataServices/dynamoDbUtil.ts
carvantes: updated.

truthy is fine here. empty string is not a valid tenantId. tenantId is validated on the router.

src/dataServices/dynamoDbBundleService.ts
rsmayda: Agreed cool good to know future plans :)

rsmayda (approved):

PR #87, opened 2021-06-18
feat: Add multi tenancy to S3

Bingjiling:
Issue #, if available:

Description of changes:

* Update S3 to support multi-tenancy 
* Update unit test 

### DIFF
--- a/src/objectStorageService/s3DataService.test.ts
+++ b/src/objectStorageService/s3DataService.test.ts
@@ -22,6 +22,8 @@ import validV3JpegBinary from '../../sampleData/validV3JpegBinary.json';
 import DynamoDbDataService from '../dataServices/__mocks__/dynamoDbDataService';
 import { S3DataService } from './s3DataService';
 
+import S3ObjectStorageService from './s3ObjectStorageService';
+
 jest.mock('./s3ObjectStorageService');
 
 describe('SUCCESS CASES: Testing create, read, update, delete of resources; version 4; multi-tenancy enabled', () => {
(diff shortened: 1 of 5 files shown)


src/objectStorageService/s3DataService.ts
```diff
@@ -109,6 +129,7 @@ export class S3DataService implements Persistence {
     }
 
     async deleteResource(request: DeleteResourceRequest) {
+        this.assertValidTenancyMode(request.tenantId);
         await this.dbPersistenceService.readResource(request);
         await S3ObjectStorageService.deleteBasedOnPrefix(request.id);
```
carvantes: delete should consider the `tenantId`

carvantes (changes requested):
Overall looks good, but I think changes are needed for the delete operation.

src/objectStorageService/s3DataService.ts
Bingjiling: Good catch! Updating now...

src/objectStorageService/s3DataService.test.ts
```diff
@@ -49,6 +51,131 @@ describe('SUCCESS CASES: Testing create, read, update, delete of resources; vers
...
+                resourceType: 'Binary',
+                contentType: 'application/pdf',
+                presignedPutUrl: 'https://VALID_S3_PUT_URL.com/tenant1/id_1.pdf/VALID_TEMP_CREDENTIAL',
+            },
+        });
+
+        expect(response.resource.data).toBeUndefined();
+        expect(response.resource.content).toBeUndefined();
+        expect(response.resource.id).toBeDefined();
+        expect(response.resource.meta).toBeDefined();
```
rsmayda, line 78: Nit: may be able to shorten your tests by use [strictEqual](https://jestjs.io/docs/expect#tostrictequalvalue)

carvantes (approved):

PR #88, opened 2021-06-23
feat: Adding support for versioned reference links

justinusmenzel:
Description of changes:
* Added `versionedLinks` as optional constructor argument. It allows for fine tuned configuration of when to use versioned reference links. 
* For each bundle resource, check if resource type is a key in `versionedLinks`. If yes, check for each reference in resource if path is in `versionedLinks` value. If reference link does not already point to a version of the target resource, determine version and update reference link.


src/dataServices/dynamoDbBundleService.ts
```diff
@@ -37,16 +42,24 @@ export class DynamoDbBundleService implements Bundle {
 
     private dynamoDb: DynamoDB;
 
-    private maxExecutionTimeMs: number;
+    private readonly maxExecutionTimeMs: number;
 
     private static readonly dynamoDbMaxBatchSize = 25;
 
+    private readonly versionedLinks: any;
```
rsmayda: Instead of any be more precise

```suggestion
    private readonly versionedLinks: Record<string, string[]>;
```

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -37,16 +42,24 @@ export class DynamoDbBundleService implements Bundle {
...
+
     // Allow Mocking DDB
-    constructor(dynamoDb: DynamoDB, supportUpdateCreate: boolean = false, maxExecutionTimeMs?: number) {
+    constructor(
+        dynamoDb: DynamoDB,
+        supportUpdateCreate: boolean = false,
+        maxExecutionTimeMs?: number,
+        versionedLinks?: Record<string, string[]>,
```
rsmayda: Can you add comments to this constructor now versionedLinks isn't clear what it is

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -270,6 +286,81 @@ export class DynamoDbBundleService implements Bundle {
         }
     }
 
+    private async updatedReferences(requests: BatchReadWriteRequest[], lockedItems: ItemRequest[]) {
+        const idToVersionId: Record<string, string> = {};
+        lockedItems.forEach((itemRequest: ItemRequest) => {
+            if (itemRequest.operation === 'update' && itemRequest.vid) {
```
rsmayda, line 330: Is this `vid` coming from the bundle request itself or is it pull from dynamo in the `lockItems` call?

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -270,6 +286,81 @@ export class DynamoDbBundleService implements Bundle {
...
+                if (references.length === 0) {
+                    return;
+                }
+                const versionedLinksSet = new Set(versionedLinksArray);
+                await Promise.all(
+                    references.map(async (referencePath: string) => {
+                        if (!versionedLinksSet.has(referencePath)) {
+                            return;
+                        }
+                        let entryReference = flattenedResources[referencePath];
```
rsmayda: nit: more explicit name
```suggestion
                        let entryReferenceValue = flattenedResources[referencePath];
```

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -270,6 +286,81 @@ export class DynamoDbBundleService implements Bundle {
...
+                                try {
+                                    const itemResponse: GenericResponse = await this.dynamoDbHelper.getMostRecentResource(
+                                        resourceType,
+                                        id,
+                                        projectionExpression,
+                                    );
+                                    const { meta } = itemResponse.resource;
+                                    vid = meta.versionId;
+                                } catch (e) {
+                                    logger.error('Failed to find most recent resource', e);
```
rsmayda: what should be do in this case? swallow the error and move on? or fail the transaction?

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -24,6 +28,7 @@ import DynamoDbHelper from './dynamoDbHelper';
...
...lProduct|MedicinalProductAuthorization|MedicinalProductContraindication|MedicinalProductIndication|MedicinalProductIngredient|MedicinalProductInteraction|MedicinalProductManufactured|MedicinalProductPackaged|MedicinalProductPharmaceutical|MedicinalProductUndesirableEffect|MessageDefinition|MessageHeader|MolecularSequence|NamingSystem|NutritionOrder|Observation|ObservationDefinition|OperationDefinition|OperationOutcome|Organization|OrganizationAffiliation|Patient|PaymentNotice|PaymentReconciliation|Person|PlanDefinition|Practitioner|PractitionerRole|Procedure|Provenance|Questionnaire|QuestionnaireResponse|RelatedPerson|RequestGroup|ResearchDefinition|ResearchElementDefinition|ResearchStudy|ResearchSubject|RiskAssessment|RiskEvidenceSynthesis|Schedule|SearchParameter|ServiceRequest|Slot|Specimen|SpecimenDefinition|StructureDefinition|StructureMap|Subscription|Substance|SubstanceNucleicAcid|SubstancePolymer|SubstanceProtein|SubstanceReferenceInformation|SubstanceSourceMaterial|SubstanceSpecification|SupplyDelivery|SupplyRequest|Task|TerminologyCapabilities|TestReport|TestScript|ValueSet|VerificationResult|VisionPrescription)\/([A-Za-z0-9\-.]{1,64})(\/_history\/[A-Za-z0-9\-.]{1,64})?/;
```
rsmayda: Can we move this to a `regExpression.ts` file? (similar to the `routing` [one](https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/regExpressions.ts)). Also could you steal the tests from the routing package https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/regExpressions.test.ts#L8

rsmayda (commented):
Could you also explain how you tested this new code?

src/dataServices/dynamoDbBundleService.ts
carvantes: Add the new constructor param as:
```suggestion
        { versionedLinks }: { versionedLinks?: Record<string, string[]> } = {},
```
Reasons for this are explained on https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/86#discussion_r656733648

src/dataServices/dynamoDbBundleService.ts
justinusmenzel: The `vid` is pulled from dynamo in the `lockItems` call

src/dataServices/dynamoDbBundleService.ts
justinusmenzel: Good question. Nobody likes silent errors. It might hint at a programming error if those occur. It's probably safer to fail the tx.

justinusmenzel:
> Could you also explain how you tested this new code?

I haven't tested it out yet. I just wanted to get some general feedback to see if the approach makes sense. Then add a better unit test and then test this out in an integration test setting.

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -176,6 +228,9 @@ export class DynamoDbBundleService implements Bundle {
         const itemReadPromises = itemsToLock.map(async itemToLock => {
             const projectionExpression = 'id, resourceType, meta';
             try {
+                logger.info(
```
carvantes, line 231: consider swirthing to logger.debug

carvantes: `logger.debug` seems more appropiate here

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -270,6 +325,98 @@ export class DynamoDbBundleService implements Bundle {
...
+            }
+        });
+
+        await Promise.all(
+            createOrUpdates.map(async (request: BatchReadWriteRequest) => {
+                const { resource } = request;
+                const versionedLinksArray = this.versionedLinks ? this.versionedLinks[request.resourceType] : [];
+                if (!versionedLinksArray) {
+                    return;
+                }
```
carvantes: Not sure if this is correct.

I think that the intention is to return early if `versionedLinksArray` is not present, but an empty array is a truthy value

maybe try:
```
const versionedLinksArray = this?.versionedLinks?.[request.resourceType];
if (!versionedLinksArray) {
    return;
}
```

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -270,6 +325,98 @@ export class DynamoDbBundleService implements Bundle {
...
+            // Setting version id to '1' of resources in the bundle that have not been locked. because
+            // if updateCreateSupported==true creates may come disguised as updates. During locking they obviously weren't found
+            // now we don't want to search for them again and then fail because we can't find them.
+            if (request.operation === 'update' && this.updateCreateSupported && !(key in idToVersionId)) {
+                idToVersionId[key] = '1';
+            }
+        });
+
+        await Promise.all(
+            createOrUpdates.map(async (request: BatchReadWriteRequest) => {
```
carvantes: I think that this section can be simplified.

The code is hard to follow since there are nested `Promise.all` that are mostly just executing synchronous code. The only truly async part is the call to DDB.

I suggest to refactor to do the synchronous stuff first with regular for loops or array functions, and only use `Promise.all` for the call to DDB

Maybe something like:
```js
const requestsWithReferencesThatMustBeLookedUp = createOrUpdates
    .filter(resourceTypeInVersionedLinksArray)
    .flatMap(extractReferences)
    .filter(referenceInVersionedLinksArray);

// requestsWithReferencesThatMustBeLookedUp is something like
// [
//     {
//         request: '<original request object>',
//         reference: {
//             path: 'a.b.c',
//             value: 'val'
//         },
//     },
//     ...
// ]

const requestsForDDB = [];

for (const requestObj of requestsWithReferencesThatMustBeLookedUp) {
    vid = findInIdToVersionId(requestObj)
    
    if (vid){
        // set reference to `/_history/${vid}`
    } else {
        requestsForDDB.push(requestObj);
    }
}

const ddbResults = await Promise.all(requestsForDDB)

for (const ddbResult of ddbResults) {
    // set reference to `/_history/${vid}`
}
```

src/dataServices/dynamoDbBundleService.ts
justinusmenzel: Yeah, that's a left over from debugging and setting up the test. I'll remove it.

src/dataServices/dynamoDbBundleService.ts
justinusmenzel: These are great suggestions. Thank you.

carvantes (dismissed):

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -37,16 +42,38 @@ export class DynamoDbBundleService implements Bundle {
...
+     * @param supportUpdateCreate
+     * @param maxExecutionTimeMs
+     * @param versionedLinks Data structure to control for which resourceTypes (key) which references (array of paths) should be modified,
+     * so that they point to the current (point in time) version of the referenced resource.
+     * For example:
+     *  {
+     *      "ExplanationOfBenefit": [ "careTeam.reference" ]
+     *  }
+     * says: for resources of type ExplanationOfBenefit, make sure the careTeam.reference url points to the current
+     * version of the practitioner resource.
```
rsmayda:
```suggestion
     * says: for resource type ExplanationOfBenefit, make sure the careTeam.reference url points to the current
     * version of the practitioner resource. 
     * For example, that would mean the reference would be corrected to: `{fhirURl}/Practitioner/1234/_history/<vid>`; instead of the default: `{fhirURl}/Practitioner/1234`
```

src/dataServices/dynamoDbBundleService.test.ts
```diff
@@ -501,4 +565,267 @@ describe('atomicallyReadWriteResources', () => {
...
+                id,
+                resourceType: 'Patient',
+                name: [
+                    {
+                        family: 'Jameson',
+                        given: ['Matt'],
+                    },
+                ],
+                managingOrganization: {
+                    reference: 'Organization/org1',
```
rsmayda: Could we have 1 test case of an absolute URL? ie:
```suggestion
                    reference: '${apiUrl}/Organization/org1',

```

rsmayda (changes requested):

rsmayda (approved):

PR #90, opened 2021-06-25
feat: Multi tenancy bulk export

Bingjiling:
Issue #, if available:

Description of changes:

* Update bulk export DDB and jobs to accommodate tenantId
* Create/update unit tests for ddbDataService, ddbParamBuilder and export jobs 

### DIFF
--- a/src/bulkExport/getJobStatus.test.ts
+++ b/src/bulkExport/getJobStatus.test.ts
@@ -6,9 +6,11 @@
 import * as AWSMock from 'aws-sdk-mock';
 import AWS from 'aws-sdk';
 import { QueryInput } from 'aws-sdk/clients/dynamodb';
+import sinon from 'sinon';
 import { getJobStatusHandler } from './getJobStatus';
 import { BulkExportStateMachineGlobalParameters } from './types';
 import { DynamoDBConverter } from '../dataServices/dynamoDb';
+import DynamoDbParamBuilder from '../dataServices/dynamoDbParamBuilder';
 
 AWSMock.setSDKInstance(AWS);
 
@@ -53,6 +55,48 @@ describe('getJobStatus', () => {
                 isCanceled: false,
             },
         });
+    });
+
+    test('completed job in multi-tenancy mode', async () => {
+        const event: BulkExportStateMachineGlobalParameters = {
+            jobId: '1',
+            tenantId: 'tenant1',
+            exportType: 'system',
+            transactionTime: '',
+            executionParameters: {
+                glueJobRunId: 'jr_1',
+            },
+        };
+        process.env.GLUE_JOB_NAME = 'jobName';
+        AWSMock.mock('Glue', 'getJobRun', (params: any, callback: Function) => {
+            callback(null, {
+                JobRun: {
+                    JobRunState: 'SUCCEEDED',
+                },
+            });
+        });
+        const getItemSpy = sinon.spy();
+        AWSMock.mock('DynamoDB', 'getItem', (params: QueryInput, callback: Function) => {
+            getItemSpy(params);
+            callback(null, {
+                Item: DynamoDBConverter.marshall({
+                    jobId: 'tenan1|2a937fe2-8bb1-442b-b9be-434c94f30e15',
+                    jobStatus: 'in-progress',
+                }),
+            });
+        });
+        await expect(getJobStatusHandler(event, null as any, null as any)).resolves.toEqual({
+            jobId: '1',
+            tenantId: 'tenant1',
+            exportType: 'system',
+            transactionTime: '',
+            executionParameters: {
+                glueJobRunId: 'jr_1',
+                glueJobRunStatus: 'SUCCEEDED',
+                isCanceled: false,
+            },
+        });
+        expect(getItemSpy.getCall(0).args[0]).toMatchObject(DynamoDbParamBuilder.buildGetExportRequestJob('tenant1|1'));
     });
 
     test('failed job', async () => {
--- a/src/bulkExport/stopExportJob.ts
+++ b/src/bulkExport/stopExportJob.ts
@@ -10,7 +10,10 @@ import getComponentLogger from '../loggerBuilder';
 
 const logger = getComponentLogger();
 
-export const stopExportJobHandler: Handler<BulkExportStateMachineGlobalParameters, { jobId: string }> = async event => {
+export const stopExportJobHandler: Handler<
+    BulkExportStateMachineGlobalParameters,
+    BulkExportStateMachineGlobalParameters
+> = async event => {
     const { GLUE_JOB_NAME } = process.env;
     if (GLUE_JOB_NAME === undefined) {
         throw new Error('GLUE_JOB_NAME environment variable is not defined');
@@ -31,7 +34,5 @@ export const stopExportJobHandler: Handler<BulkExportStateMachineGlobalParameter
         logger.error('Failed to stop job', JSON.stringify(stopJobRunResponse));
         throw new Error(`Failed to stop job ${glueJobRunId}`);
     }
-    return {
-        jobId: event.jobId,
-    };
+    return event;
 };
--- a/src/dataServices/dynamoDbDataService.ts
+++ b/src/dataServices/dynamoDbDataService.ts
@@ -39,22 +39,6 @@ import DynamoDbHelper from './dynamoDbHelper';
 import { getBulkExportResults, startJobExecution } from '../bulkExport/bulkExport';
 import { BulkExportJob } from '../bulkExport/types';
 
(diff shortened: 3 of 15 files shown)


src/bulkExport/stopExportJob.ts
```diff
@@ -31,7 +34,5 @@ export const stopExportJobHandler: Handler<BulkExportStateMachineGlobalParameter
         logger.error('Failed to stop job', JSON.stringify(stopJobRunResponse));
         throw new Error(`Failed to stop job ${glueJobRunId}`);
     }
-    return {
-        jobId: event.jobId,
-    };
+    return event;
```
rsmayda, line 37: It doesn't look like we are modifying `event` at all do we need to return it?

src/dataServices/dynamoDbDataService.ts
```diff
@@ -325,6 +310,30 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
+            type: initiateExportRequest.type ?? '',
+            transactionTime: initiateExportRequest.transactionTime,
+            jobStatus: initialStatus,
+            jobFailedMessage: '',
+        };
+        if (this.enableMultiTenancy) {
+            exportJob = {
+                ...exportJob,
+                tenantId: initiateExportRequest.tenantId,
+            };
```
rsmayda: I don't think you need to rebuild it

```suggestion
            exportJob.tenantId = initiateExportRequest.tenantId;
```

src/dataServices/dynamoDbDataService.ts
```diff
@@ -325,6 +310,30 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
         return getExportStatusResponse;
     }
 
+    buildExportJob(initiateExportRequest: InitiateExportRequest): BulkExportJob {
```
rsmayda, line 324: Why make this a class method?

src/dataServices/dynamoDbParamBuilder.ts
```diff
@@ -142,9 +144,16 @@ export default class DynamoDbParamBuilder {
     }
 
     static buildPutCreateExportRequest(bulkExportJob: BulkExportJob) {
+        const newItem: any = { ...bulkExportJob };
+        if (newItem.tenantId !== undefined) {
```
rsmayda: Could use "falsy"?

```suggestion
        if (!newItem.tenantId) {

```

src/dataServices/dynamoDbParamBuilder.ts
```diff
@@ -166,28 +175,28 @@ export default class DynamoDbParamBuilder {
         return params;
     }
 
-    static buildUpdateExportRequestJobStatus(jobId: string, jobStatus: ExportJobStatus) {
+    static buildUpdateExportRequestJobStatus(jobId: string, jobStatus: ExportJobStatus, tenantId?: string) {
         const params = {
             TableName: EXPORT_REQUEST_TABLE,
             Key: DynamoDBConverter.marshall({
-                jobId,
+                jobId: buildHashKey(jobId, tenantId),
```
rsmayda: Could pull out `buildHashKey(jobId, tenantId)` call into a variable and use it in both places:
```ts
const jobId = buildHashKey(jobId, tenantId)
```

src/bulkExport/stopExportJob.ts
Bingjiling: Yes, we need to return it as the input for the next step in the state machine for DDB status update. Before multi-tenancy, the next step only needs jobId, now it needs both tenantId and jobId. Passing tenantId as `tenantId: event.tenantId` would error out the state machine if event does not have tenantId attribute, so I'm passing the event altogether.

src/dataServices/dynamoDbDataService.ts
Bingjiling: The method uses class variable `this.enableMultiTenancy`. Although it is the same as checking if tenantId exist, I think it's more readable using `this.enableMultiTenancy`.

src/dataServices/dynamoDbDataService.ts
```diff
@@ -265,8 +250,8 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
         return allJobStatusItems;
     }
 
-    async cancelExport(jobId: string): Promise<void> {
-        const jobDetailsParam = DynamoDbParamBuilder.buildGetExportRequestJob(jobId);
+    async cancelExport(jobId: string, tenantId?: string): Promise<void> {
+        const jobDetailsParam = DynamoDbParamBuilder.buildGetExportRequestJob(jobId, tenantId);
```
carvantes: add `this.assertValidTenancyMode` here

src/dataServices/dynamoDbDataService.ts
```diff
@@ -280,12 +265,12 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
 
-        const params = DynamoDbParamBuilder.buildUpdateExportRequestJobStatus(jobId, 'canceling');
+        const params = DynamoDbParamBuilder.buildUpdateExportRequestJobStatus(jobId, 'canceling', tenantId);
         await this.dynamoDb.updateItem(params).promise();
     }
 
-    async getExportStatus(jobId: string): Promise<GetExportStatusResponse> {
-        const jobDetailsParam = DynamoDbParamBuilder.buildGetExportRequestJob(jobId);
+    async getExportStatus(jobId: string, tenantId?: string): Promise<GetExportStatusResponse> {
+        const jobDetailsParam = DynamoDbParamBuilder.buildGetExportRequestJob(jobId, tenantId);
```
carvantes: add this.assertValidTenancyMode here

src/dataServices/dynamoDbDataService.ts
```diff
@@ -217,9 +201,10 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
     }
 
     async initiateExport(initiateExportRequest: InitiateExportRequest): Promise<string> {
+        this.assertValidTenancyMode(initiateExportRequest.tenantId);
         await this.throttleExportRequestsIfNeeded(initiateExportRequest.requesterUserId);
```
carvantes: The `throttleExportRequestsIfNeeded` needs to account for multi-tenancy. I missed this on the design doc. 

The basic idea is that each tenant gets their own allowed `MAXIMUM_SYSTEM_LEVEL_CONCURRENT_REQUESTS` instead of all of them sharing the same requests count. Otherwise "noisy neighbors" can prevent small tenants from starting export requests.

I *think* that this is not a big change

src/dataServices/dynamoDbParamBuilder.ts
```diff
@@ -142,9 +144,16 @@ export default class DynamoDbParamBuilder {
     }
 
     static buildPutCreateExportRequest(bulkExportJob: BulkExportJob) {
+        const newItem: any = { ...bulkExportJob };
+        if (newItem.tenantId) {
+            newItem[TENANT_ID_FIELD] = newItem.tenantId;
```
carvantes: There's no need to use `TENANT_ID_FIELD='_tenantId'` here.

_tenantId is prefixed with an underscore because on the resources table we prefix with underscore all "internal"fields to avoid conflicts with FHIR resources fields. 

The bulk export jobs ddb records are entirely our own so we don't need "internal" fields.

simply naming the field `tenantId` makes more sense

src/bulkExport/getJobStatus.test.ts
```diff
@@ -6,9 +6,11 @@
 import * as AWSMock from 'aws-sdk-mock';
 import AWS from 'aws-sdk';
 import { QueryInput } from 'aws-sdk/clients/dynamodb';
+import sinon from 'sinon';
```
carvantes, line 9: nit: It's weird to use `sinon` instead of `jest.fn()` given that we are using jest. Mixing test libraries adds complexity to tests and jest alone is sufficient for these tests (I think).

I know other tests in this package use `sinon` but I'm not sure if that was needed to begin with.

No need to rewrite tests. Just something to keep in mind.

https://jestjs.io/docs/expect#tohavebeencalledwitharg1-arg2-
https://jestjs.io/docs/mock-function-api#mockfnmockcalls

src/dataServices/dynamoDbDataService.ts
Bingjiling: Yes, it makes sense and it's not a big change. I'll make the update.

src/bulkExport/getJobStatus.test.ts
Bingjiling: Yep, that makes sense. I'll try avoid using sinon going forward.

carvantes (approved):

PR #91, opened 2021-06-28
feat: update ddbToEs to handle multi-tenant records

carvantes:
When records in ddb have a `_tenantId` the following happens in the ddbToEs lambda:

- Minor reformatting of the DDB record before indexing in ES. `_id` must be removed and `id` is reverted to the FHIR resource id instead of being the DDB composite key.
- Write to ES using a tenant-specific alias.

### DIFF
--- a/src/ddbToEs/ddbToEsHelper.ts
+++ b/src/ddbToEs/ddbToEsHelper.ts
@@ -92,6 +92,10 @@ export default class DdbToEsHelper {
                                     type: 'keyword',
                                     index: true,
                                 },
+                                _tenantId: {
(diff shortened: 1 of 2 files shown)


rsmayda (approved):

rsmayda:
Should we set _tenant_id as a keyword type?

carvantes:
> Should we set _tenant_id as a keyword type?

yeah, that's probably better. I was thinking on letting dynamic mapping take care to make it backwards compatible, but multi-tenancy is a breaking change for several other reasons so it makes sense to add it to our mapping.

I'll make the change.

PR #94, opened 2021-07-16
fix: list bulk export files with tenant prefix

carvantes:
Export results list was empty since the `tenantId` was not being used as prefix in calls to `s3.listObjects`

### DIFF
--- a/src/bulkExport/bulkExport.test.ts
+++ b/src/bulkExport/bulkExport.test.ts
@@ -38,6 +38,19 @@ describe('getBulkExportResults', () => {
         ]);
     });
 
+    test('happy case with tenantId', async () => {
+        AWSMock.mock('S3', 'listObjectsV2', (params: any, callback: Function) => {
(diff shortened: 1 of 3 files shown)


src/bulkExport/bulkExport.test.ts
```diff
@@ -38,6 +38,19 @@ describe('getBulkExportResults', () => {
         ]);
     });
 
+    test('happy case with tenantId', async () => {
+        AWSMock.mock('S3', 'listObjectsV2', (params: any, callback: Function) => {
+            callback(null, {
```
Bingjiling, line 44: Check here the prefix passed in is as expected?

Bingjiling (approved):
LGTM

src/bulkExport/bulkExport.test.ts
carvantes: makes sense. I added the check before merging

Issue #93, opened 2021-07-15
[Bug] Synchronization of DDB data with ES using the DDB-to-ES lambda is slow


arnabguptadev:
**Describe the bug**
The synchronization of DynamoDB data to ElasticSearch using the DynamoDB stream is slow and results in large number of errors when the data volume is high.

**To Reproduce**
1. Load a large amount of data in DynamoDB
2. Check the aws cloudwatch console to monitor ElasticSearch
3. The synchronization goes on for a large time with large number of errors.

**Expected behavior**
The synchronization should complete within acceptable time and reduced errors.

**Screenshots**

**Versions (please complete the following information):**
 - 3.4.0

**Additional context**
As discussed with AWS team, using batch update APIs for ElasticSearch may significantly address this issue.

carvantes:
The fix is merged. 

We'll close this issue once we release a new version of this package

arnabguptadev:
Thanks @carvantes! Appreciate it. Will test and ask team to close.

rsmayda:
Merged in as 3.6.1

Issue #96, opened 2021-07-21
[Feature Request] Need ability to specify index/alias name in handleDdbToEsEvent


arnabguptadev:
**Is your feature request related to a problem? Please describe.**
Need a way to populate ElasticSearch index without using the standard alias used in AWS Fhirworks DDB to ES Lambda. Since the lambda may fail, we need ways to populate the index with data or even want to rebuild specific indices if needed. Currently the code to do this - [handleDdbToEsEvent](https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/42df1e44c742441142dbde98b2fb8d51d6cf3c0b/src/ddbToEs/index.ts#L16) - makes an assumption on the alias name to use. This alias is backed by an index. We want to be able to populate that index independent of the current index that backs the alias. That way, we can reindex and then switch the alias without a downtime. We plan to use the above method - but the assumption made in the code always forces the data to go through a fixed alias name.

**Describe the solution you'd like**
There should be a way to specify the target index in the method handleDdbToEsEvent. This can either be a part of the entries in the event object that is a method parameter or a callback function - that - given a resourceType - can return the index to use.

**Describe alternatives you've considered**
We currently use our code to batch the updates. We do want to use the AWS method once it supports batch.

**Additional context**

SanketD92:
Thank you for opening this feature request, @arnabguptadev. We have added this to our team's backlog and will work on it based on priority.

PR #97, opened 2021-07-26
fix: change ES writes to use Bulk APIs

rsmayda:
Issue #, if available: #93

Description of changes:
- Change to use ES bulk apis
- When creating new indices/aliases:
  - only create a new index if and only if we can't find the alias either
  - create only a new alias if we can find an index with the <resourcetype> as the name
- In `index.ts` we are caching the aliases we know of in `knownAliases`
- Previously we talked about letting the `bulk` fail if it can't find an alias, but with the bulk API and using upserts it doesn't fail and will actually then start creating indices with their name being `<resourceType>-alias`
  -  note we are doing bulk reads to determine if alias exists.


src/ddbToEs/ESBulkCommand.ts
```diff
@@ -0,0 +1,15 @@
...
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+export type OperationType = 'delete' | 'upsert-AVAILABLE' | 'upsert-DELETED';
+
+export default interface ESBulkCommand {
+    bulkCommand: any[];
+    /**
+     * This will be a unquie identifier or composite id i.e. `<id>_<vid>`
```
carvantes: typo
```suggestion
     * This will be a unique identifier or composite id i.e. `<id>_<vid>`
```

src/ddbToEs/ddbToEsHelper.ts
```diff
@@ -42,57 +43,89 @@ export default class DdbToEsHelper {
...
-                                    index: true,
-                                },
-                                documentStatus: {
-                                    type: 'keyword',
-                                    index: true,
+            const promises: any[] = [];
+            Array.from(indicesToCreate).forEach((index: string) => {
+                const alias = `${index}-alias`;
+                // Only create index when we also need to create an alias
+                if (aliasesToCreate.has(alias)) {
```
carvantes, line 81: This `if` seems redundant (can the condition ever be false?). Given the way `indicesToCreate` is constructed, there's no way that alias already exists and the correspondin index does not.

src/ddbToEs/ddbToEsHelper.ts
```diff
@@ -42,57 +43,89 @@ export default class DdbToEsHelper {
...
+        if (allFound) {
+            // All needed aliases exist
+            return;
+        }
+
+        logger.debug('There are missing aliases');
+
+        const indicesToCreate: Set<string> = new Set();
+        listOfAliases.forEach((alias: string) => {
+            indicesToCreate.add(alias.substring(0, alias.length - 6)); // remove '-alias'
```
carvantes: I feel that there's a bit too much string manipulation to come up with the index name and aliases. Very soon we'll need to update this code to account for the multi-tenant aliases (that follow a different naming convention) so it'd be nice to DRY index/alias naming 

My recommendation would be to update this method to accept resourceTypes instead of aliases:
```
createIndexAndAliasIfNotExist(resourceTypes: Set<string>)
```
I think it's clearer to derive the index/alias name from the resource type than trying to get the index name by removing chars from the alias.

Also, I suggest adding a function:
```
const getAliasName = (resourceType: string): string
``` 
to be the single place where the alias naming convention is defined instead of having string concats across the codebase(i.e. `${lowercaseResourceType}-alias`)
https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/79c77f878bfb5eb17cafe82a28c24c5b340ecdfa/src/ddbToEs/ddbToEsHelper.ts#L22



soon we'll need to update this code to handle the multi-tenant aliases as well

src/ddbToEs/ddbToEsHelper.ts
rsmayda: The scenario I am trying to avoid is if/when index corruption happens and a new index is name something non-traditional `patient-new` in this situation we would have `patient` in our index set but `patient-alias` isn't in the alias set, so we shouldn't create `patient` index. 

The assumption I am making here is FWoA really only cares if the Alias is set up and connected to *a nonspecific* index

src/ddbToEs/ddbToEsHelper.ts
```diff
@@ -42,57 +41,88 @@ export default class DdbToEsHelper {
...
                 }
-            }
+            });
+
+            Array.from(aliasesToCreate).forEach((alias: string) => {
+                // Create Alias
+                logger.info(`create alias ${alias}`);
+                promises.push(
+                    this.ElasticSearch.indices.putAlias({
+                        index: alias.substring(0, alias.length - 6),
```
carvantes: can we get rid of this substring operation as well?

rsmayda: Sadly no but we can move it into a method -- let me do that

src/ddbToEs/ddbToEsHelper.ts
carvantes: I see. I wasn't considering index creation outside of this lambda. That's fair.

src/ddbToEs/ddbToEsHelper.ts
carvantes: I think that it can be done, but it requires some additional refactoring. It seems that keeping a Set of aliases disconnected from the corresponding index names makes this harder.

Maybe it was better to start off with a data structure like:
```
[
 {
  resourceType: string,
  index: string,
  alias: string,
 },
...
]
```
To explicitly keep the relations between index and alias.

There's also a feature request from Abacus to use custom alias names: https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/issues/96, that would make it harder it harder to derive an index from the alias name with string manipulation(custom alias can be anything)

I think that to accomodate the feature request and the multi-tenancy aliases we'll need to eventually do this kind to refactoring, but the current approach works fine for now.

carvantes (approved):

src/ddbToEs/ddbToEsHelper.ts
rsmayda: Good call out, makes sense

PR #98, opened 2021-07-28
feat: persistence change for group export

Bingjiling:
Issue #, if available:

Description of changes:

* Update to persistence to pass in parameters needed for group export
* Reject request if user ask for resource type outside of allowed types
* Clean up export request before saving to DDB
* Unit test 

### DIFF
--- a/src/bulkExport/bulkExport.ts
+++ b/src/bulkExport/bulkExport.ts
@@ -73,7 +73,18 @@ export const getBulkExportResults = async (
 };
 
 export const startJobExecution = async (bulkExportJob: BulkExportJob): Promise<void> => {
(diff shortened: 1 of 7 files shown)


rsmayda (commented):
Overall looks good! For the glue script could we use a set for 'types' instead of a list? Not sure if that part has already been handled

Bingjiling:
> Overall looks good! For the glue script could we use a set for 'types' instead of a list? Not sure if that part has already been handled

Yes! Types are used as a set in the Glue script, it's handled inside the script [here](https://github.com/awslabs/fhir-works-on-aws-deployment/pull/389/files#diff-1f14038f9237de4f90e34192aa8c90103e82ece6314077b567c3cb47c3ce6dfcR166)

carvantes (approved):

PR #102, opened 2021-08-11
fix: reroute locking errors from server errors to conflict errors where applicable

ssvegaraju:
Issue #, if available:

Description of changes:


src/dataServices/dynamoDbBundleService.ts
```diff
@@ -314,9 +315,20 @@ export class DynamoDbBundleService implements Bundle {
             });
         } catch (e) {
             logger.error('Failed to lock', e);
+            for (let i = 0; i < lockedItems.length; i += 1) {
+                // if the request in the bundle was a PUT request, we have a race condition causing conflicts when locking/unlocking resources
+                if (lockedItems[i].isOriginalUpdateItem) {
+                    throw new ResourceConflictError(
```
Bingjiling: We can not be sure which item actually caused the conflict error here, there could be multiple update entries and only one of them is causing the conflict issue. Also, the Transaction method will unlock any successful locked items if the lock method return error information in a resolved promise [here](https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/102/files#diff-bad4ab4404b21b69f94c2366d2b94daea20d791e5fd77ebeef937d45efa3b9c6R102), throwing an error will abort the Transaction method before it can unlock items.  

I think the following block(line 329 - 335) already covers this situation right?

src/dataServices/dynamoDbBundleService.test.ts
```diff
@@ -91,7 +91,7 @@ describe('atomicallyReadWriteResources', () => {
                 success: false,
                 message: 'Failed to lock resources for transaction. Please try again after 35 seconds.',
                 batchReadWriteResponses: [],
-                errorType: 'SYSTEM_ERROR',
+                errorType: 'USER_ERROR',
```
Bingjiling: Can we create another test case where update transaction conflict occurs?

src/dataServices/dynamoDbBundleService.ts
ssvegaraju: Updated the logic to use the errors returned from the DynamoDB call to discern when a conflict error occurs instead of this implementation

src/dataServices/dynamoDbBundleService.test.ts
ssvegaraju: Restored this test case, created a script to test for transaction conflicts separately.

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -335,6 +335,16 @@ export class DynamoDbBundleService implements Bundle {
             });
         } catch (e) {
             logger.error('Failed to lock', e);
+            const err = JSON.parse(JSON.stringify(e));
```
rsmayda: Do we need to do this deserialization and the serialization again?

src/dataServices/dynamoDbBundleService.ts
```diff
@@ -523,6 +533,10 @@ export class DynamoDbBundleService implements Bundle {
 
         const newLockedItems = this.removeLocksFromArray(lockedItems, itemsToRemoveFromLock);
 
+        // if batchReadWriteEntryResponses is empty, don't throw error here:
```
rsmayda:
```suggestion
        // if batchReadWriteEntryResponses is empty, don't throw error here, since there is nothing to rollback
```

src/dataServices/dynamoDbBundleService.ts
ssvegaraju: Turns out we don't need it, VSCode Intellisense just likes to complain about `e` being unknown. I've removed this and it passes linting + tests.

src/dataServices/dynamoDbBundleService.ts
ssvegaraju: Updated this comment!

rsmayda (approved):

PR #105, opened 2021-08-16
merge ddbToEs code

carvantes:
The merge from mainline was tricky since https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/commit/7eee06e5955bb980abd17dced35a86236ea10189 did a big refactor on the way indices/aliases are created that conflicted with how the per tenant aliases were setup.

This is the result of the merge.

You may want to focus on https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/commit/1fa12c645b749f66fc5a1004ed5060d3cf241b56 (I just realized that you can leave comments in commits), That's the commit that "fixes" the merge result to account for multi-tenant alias. One important thing is that in multi-tenancy a single index can have several aliases (one per tenant), so  a some refactoring was needed.


src/ddbToEs/ddbToEsHelper.ts
```diff
@@ -66,16 +58,56 @@ export default class DdbToEsHelper {
...
+        const { body: allFound } = await this.ElasticSearch.indices.existsAlias({
+            name: listOfAliases,
+            expand_wildcards: 'all',
+        });
+        if (allFound) {
+            // All needed aliases exist
+            return;
+        }
+
+        logger.debug('There are missing aliases');
```
rsmayda, line 79: To confirm we lazily create aliases? i.e. we have tenants: 1,2,3 

If a create Patient request comes in on tenant 1 the resulting ES cluster would have:
- index: `patient`
- alias: `patient-alias-tenant-1`

src/ddbToEs/ddbToEsHelper.test.ts
```diff
@@ -19,29 +20,91 @@ describe('DdbToEsHelper', () => {
         esMock.clearAll();
     });
 
-    describe('createIndexIfNotExist', () => {
-        test('Create index and alias for new index', async () => {
-            process.env.ENABLE_ES_HARD_DELETE = 'true';
+    describe('createIndexAndAliasIfNotExist', () => {
```
rsmayda, line 23: Are there test cases for testing the tenant alias format? (I couldn't find one) i.e. looking for the test cases that follow our tenanted format `generateAlias`

src/ddbToEs/ddbToEsHelper.ts
carvantes: that's correct

src/ddbToEs/ddbToEsHelper.test.ts
carvantes: I added a couple tests for the `generateAlias` method.

It's worth noting that `createIndexAndAliasIfNotExist` no longer decides the alias naming convention, it just takes aliases and indices names as parameters so the method doesn't have multi-tenancy specific code.

rsmayda (approved):

carvantes:
The PR is approved. I'll merge manually instead of merging via github since I need a merge commit instead of a "squash and merge"

PR #108, opened 2021-08-27
feat: add config options to DdbToEsSync

### LINKED ISSUE #96: [Feature Request] Need ability to specify index/alias name in handleDdbToEsEvent

carvantes:
Closes: https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/issues/96

Functionality for the default deployment remains unchanged

new optional config params:
- `esClient`: overriding the ES client. Probably easier than having to depend on several env variables
- `disableIndexAndAliasCreation`: disables index/alias creation. 
- `getAliasFn`: using custom alias/index names

An example using all the optional config values would be:
```ts
const ddbToEsSync = new DdbToEsSync({
        esClient: new Client({
            node: 'https://my-search-cluster.us-west-2.es.amazonaws.com',
            Connection: AmazonConnection,
            Transport: AmazonTransport,
        }),
        getAliasFn: (x: any) => ({
            alias: `${x.resourceType.toLowerCase()}-my-test-alias`,
            index: `${x.resourceType.toLowerCase()}-my-test-index`,
        }),
        disableIndexAndAliasCreation: true,
    });
```

Going forward, We suggest customers to not depend on `handleDdbToEsEvent` but rather on the new `DdbToEsSync` class. The original `handleDdbToEsEvent` was not really meant to be used other than as a lambda handler and is cumbersome to add config options to it without making breaking changes.


src/index.ts
```diff
@@ -8,6 +8,7 @@ export * from './dataServices/dynamoDbUtil';
 export { DynamoDb } from './dataServices/dynamoDb';
 export * from './objectStorageService/s3DataService';
 export { handleDdbToEsEvent } from './ddbToEs/index';
+export { DdbToEsSync } from './ddbToEs/DdbToEsSync';
```
rsmayda: Does this actually need to be exported? `DdbToEsSync` is just used by `handleDdbToEsEvent`

src/ddbToEs/index.ts
```diff
@@ -3,85 +3,13 @@
...
-                        : record.dynamodb.NewImage;
-                    return `{id: ${image.id.S}, vid: ${image.vid.N}}`;
-                },
-            ),
-        );
-
-        logger.error('Failed to update ES records', e);
-        throw e;
-    }
+    return ddbToEsSync.handleDDBStreamEvent(event);
```
rsmayda, line 14: +1 good separation

src/ddbToEs/DdbToEsSync.ts
```diff
@@ -0,0 +1,130 @@
+/*
```
rsmayda: Case of file: `src/ddbToEs/DdbToEsSync.ts` -> `src/ddbToEs/ddbToEsSync.ts`

src/ddbToEs/DdbToEsSync.test.ts
```diff
@@ -0,0 +1,290 @@
+/*
```
rsmayda: Case of file `src/ddbToEs/DdbToEsSync.test.ts` -> `src/ddbToEs/ddbToEsSync.test.ts`

src/ddbToEs/ddbToEsHelper.ts
```diff
@@ -35,27 +35,31 @@ const formatDocument = (ddbImage: any): any => {
     return ddbImage;
 };
 
-export default class DdbToEsHelper {
```
rsmayda, line 38: Do you think `DdbToEsHelper` is the right name for this class anymore? I feel like this "helper" class is doing A LOT. I am fine leaving it as is but wanted to pose the question with this re-write

src/ddbToEs/DdbToEsSync.test.ts
```diff
@@ -0,0 +1,290 @@
...
+
+const ddbHelperCreateIndexAndAliasIfNotExistMock = jest.fn();
+const ddbHelperExecuteEsCmds = jest.fn();
+
+// @ts-ignore
+DdbToEsHelper.prototype.createIndexAndAliasIfNotExist = ddbHelperCreateIndexAndAliasIfNotExistMock;
+// @ts-ignore
+DdbToEsHelper.prototype.executeEsCmds = ddbHelperExecuteEsCmds;
+
+// jest.mock('./ddbToEsHelper');
```
rsmayda: needed?

src/index.ts
carvantes: For FWoA alone, not really. But it is exported so that external customers can use it.

src/ddbToEs/DdbToEsSync.ts
carvantes: updated

src/ddbToEs/DdbToEsSync.test.ts
carvantes: updated

src/ddbToEs/DdbToEsSync.test.ts
carvantes: removed

src/ddbToEs/ddbToEsHelper.ts
carvantes: IMO "helpers" are almost always bad naming/design. They don't have a well defined purpose and very often end up growing a LOT.

Here I'm leaving the class name and most of it's content as is, since it works and the refactoring was mostly targeting the old `index.ts` file.

rsmayda (approved):

PR #115, opened 2021-09-20
chore: resolve transitive dependency on set-value

ssvegaraju:
Issue #, if available:

Description of changes: 
Addressing the Dependabot alerts regarding transitive dependencies on an unsafe version of `set-value` (<4.0.1) by adding the package to the `resolutions`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,8 @@
     "typescript": "^4.1.3"
   },
   "resolutions": {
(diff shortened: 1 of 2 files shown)


carvantes (approved):
lgtm. Also, I updated the title to be `chore:` instead of `fix:`. Dependency updates are always chore or similar. We don't want them to appear on changelog.

PR #117, opened 2021-09-21
chore: resolve transitive dependencies on ansi-regex

ssvegaraju:
Issue #, if available:

Description of changes:

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
   "resolutions": {
     "path-parse": "^1.0.7",
     "set-value": "^4.0.1",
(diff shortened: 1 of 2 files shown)


carvantes (commented):
Why are we doing this? I don't see a related dependabot issue

ssvegaraju:
> Why are we doing this? I don't see a related dependabot issue

Huh, that's weird. It looks like the dependabot alert is gone from all the repositories that had the issue. Will close this PR and double check the others to make sure they're valid.

PR #123, opened 2021-10-13
feat: support custom bulk export results URLs

carvantes:
Allows customers to provide an implementation of `bulkExportResultsUrlGenerator` on the `DynamoDbDataService` constructor to customize the location of bulk export results and the value of `requiresAccessToken` 

It is mostly a refactor. There are no functional changes with the default settings.

Related PRs (unit tests her fail until I publish and bump the interface version):
- https://github.com/awslabs/fhir-works-on-aws-interface/pull/85

### DIFF
--- a/src/bulkExport/bulkExportResultsUrlGenerator.ts
+++ b/src/bulkExport/bulkExportResultsUrlGenerator.ts
@@ -5,5 +5,8 @@
  */
 
 export interface BulkExportResultsUrlGenerator {
-    getUrls(options: { bucket: string; keys: string[] }): Promise<{ requiresAccessToken: boolean; urls: string[] }>;
+    getUrls(options: {
+        exportBucket: string;
+        s3Keys: string[];
+    }): Promise<{ requiresAccessToken: boolean; urls: string[] }>;
 }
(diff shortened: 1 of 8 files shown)


src/bulkExport/bulkExportResultsUrlGenerator.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+export interface BulkExportResultsUrlGenerator {
+    getUrls(options: { bucket: string; keys: string[] }): Promise<{ requiresAccessToken: boolean; urls: string[] }>;
```
rsmayda: Since we currently require export to be put in S3 thoughts on making the params more explicit? 
```suggestion
    getUrls(options: { exportBucket: string; s3Keys: string[] }): Promise<{ requiresAccessToken: boolean; urls: string[] }>;
```

carvantes: sounds good. I just renamed them

rsmayda (approved):

PR #131, opened 2022-01-12
feat: implement getActiveSubscriptions

ssvegaraju:
Issue #, if available:

Description of changes:
There is a linked change in the interface package that provides the necessary method: https://github.com/awslabs/fhir-works-on-aws-interface/pull/95.


src/dataServices/dynamoDbUtil.test.ts
```diff
@@ -272,3 +272,49 @@ describe('prepItemForDdbInsert', () => {
...
+    const id = '8cafa46d-08b4-4ee4-b51b-803e20ae8126';
+    const vid = 1;
+    const resource = {
+        resourceType: 'Subscription',
+        id: 'example',
+        status: 'active',
+    };
+    test('_subscriptionStatus is present', () => {
+        const actualItem = DynamoDbUtil.prepItemForDdbInsert(
+            clone(resource),
```
carvantes: there's no need to clone since the test data is exclusive to this test

src/dataServices/dynamoDbDataService.ts
```diff
@@ -385,6 +385,26 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
+        this.assertValidTenancyMode(params.tenantId);
+        let keyCondition = '_subscriptionStatus = :active';
+        if (params.tenantId) {
+            keyCondition += ` AND id beginsWith ${params.tenantId}`;
+        }
+        const subscriptionQuery = {
+            TableName: RESOURCE_TABLE,
+            IndexName: 'activeSubscriptions',
+            KeyConditionExpression: keyCondition,
+        };
```
carvantes: Did you test against DDB?

I think that this will fail since `:active` is missing in `ExpressionAttributeValues`.

See an example at: 
https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/e2ccf527eeb5fede5ac0a1b42511ec4ab08b7255/src/dataServices/dynamoDbParamBuilder.test.ts#L286-L297

carvantes (changes requested):
Also remove code coverage related changes

src/dataServices/dynamoDbDataService.ts
ssvegaraju: I changed the pattern of this function to match the others, using parambuilder and wrote more tests on each part of the implementation.

package.json
```diff
@@ -81,6 +81,7 @@
       "text",
       "html"
     ],
+    "collectCoverage": true,
```
carvantes: nit: remove since it is related to code coverage changes

src/dataServices/dynamoDbDataService.ts
```diff
@@ -385,6 +385,18 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
+    async getActiveSubscriptions(params: { tenantId?: string }): Promise<Record<string, any>[]> {
+        this.assertValidTenancyMode(params.tenantId);
+        const subscriptionQuery = DynamoDbParamBuilder.buildGetActiveSubscriptions(params.tenantId);
+        const queryResponse = await this.dynamoDb.query(subscriptionQuery).promise();
+        const subscriptions: Record<string, any>[] = [];
+        queryResponse.Items?.forEach((response) => {
+            const item = DynamoDBConverter.unmarshall(response);
+            subscriptions.push({[item.id]: item});
+        });
+        return subscriptions;
```
carvantes: consider the case where you need to paginate through the query results (i.e. `LastEvaluatedKey` is present)

Something like https://github.com/carvantes/fhir-works-on-aws-deployment/blob/c7c63417e655e66ee0d7ef5d8c84b2d50ee4f8ad/scripts/sort-key-migration.js#L82

src/dataServices/dynamoDbParamBuilder.ts
```diff
@@ -207,4 +207,20 @@ export default class DynamoDbParamBuilder {
 
         return params;
     }
+
+    static buildGetActiveSubscriptions(tenantId?: string) {
+        let keyCondition = '_subscriptionStatus = :active';
+        if (tenantId) {
+            keyCondition += ` AND id beginsWith ${tenantId}`;
```
carvantes: This is unsafe. `${tenantId}` may break the DDB syntax. You can instead do `AND id beginsWith :tenantId` and add the tenant id to `ExpressionAttributeValues`.

In general it is the best practice to always use `ExpressionAttributeValues`

carvantes (changes requested):
Did you test against DDB?

src/dataServices/dynamoDbDataService.ts
```diff
@@ -385,6 +385,25 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
 
+    async getActiveSubscriptions(params: { tenantId?: string }): Promise<Record<string, any>[]> {
+        this.assertValidTenancyMode(params.tenantId);
+        const subscriptionQuery = DynamoDbParamBuilder.buildGetActiveSubscriptions(params.tenantId);
+        let queryResponse = await this.dynamoDb.query(subscriptionQuery).promise();
+        const subscriptions: Record<string, any>[] = [];
+        do {
+            queryResponse.Items?.forEach((response) => {
+                const item = DynamoDBConverter.unmarshall(response);
+                subscriptions.push({ [item.id]: item });
```
carvantes: why index by `item.id`? I think that we want a simple array of subscriptions i.e. `subscriptions.push(item)`

src/dataServices/dynamoDbDataService.ts
```diff
@@ -385,6 +385,25 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
+            });
+            if (queryResponse.LastEvaluatedKey) {
+                subscriptionQuery.ExclusiveStartKey = queryResponse.LastEvaluatedKey;
+                // eslint-disable-next-line no-await-in-loop
+                queryResponse = await this.dynamoDb.query(subscriptionQuery).promise();
+            }
+        } while (queryResponse.LastEvaluatedKey);
```
carvantes: refactor so that `let queryResponse = await this.dynamoDb.query(subscriptionQuery).promise();` happens only once at the beginning of the do while. It's clearer that way

src/dataServices/dynamoDbParamBuilder.test.ts
```diff
@@ -445,3 +445,32 @@ describe('buildUpdateExportRequestJobStatus', () => {
...
+    };
+    test('Param without tenantId', () => {
+        const actualParam = DynamoDbParamBuilder.buildGetActiveSubscriptions();
+        expect(actualParam).toEqual(expectedParam);
+    });
+
+    test('tenantId present', () => {
+        const actualParam = DynamoDbParamBuilder.buildGetActiveSubscriptions('tenant1');
+
+        const clonedExpectedParam: any = cloneDeep(expectedParam);
```
carvantes: nit: all this cloning makes the test harder to read. reusing `expectedParam` is probably not worth it for just 2 tests.
I'd suggest declaring the whole `expectedParam` for multiTenancy or using snapshots (`expect(actualParam).toMatchInlineSnapshot`)

carvantes (changes requested):

src/dataServices/dynamoDbDataService.ts
ssvegaraju: That makes more sense than what I was doing, the return type confused me into thinking we wanted a dictionary of item ids and subscriptions. I've fixed that now

src/dataServices/dynamoDbDataService.ts
ssvegaraju: Refactored!

src/dataServices/dynamoDbParamBuilder.ts
ssvegaraju: Tested against DDB and revised the syntax

src/dataServices/dynamoDbDataService.ts
ssvegaraju: Unfortunately using the parambuilder to generate the query meant that it wasn't easy to add the ExclusiveStartKey, so I had to use the defineProperty function.

src/dataServices/dynamoDbDataService.test.ts
```diff
@@ -948,3 +948,47 @@ each(['cancelExport', 'getExportStatus']).test('%s:Unable to find job', async (t
         );
     }
 });
+
+describe('getActiveSubscriptions', async () => {
```
carvantes:
```suggestion
test('getActiveSubscriptions', async () => {
```

Make sure to check Jest output when writing new tests:
```
Returning a Promise from "describe" is not supported. Tests must be defined synchronously.
Returning a value from "describe" will fail the test in a future version of Jest.
```

src/dataServices/dynamoDbDataService.ts
```diff
@@ -385,6 +385,30 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
+                const item = DynamoDBConverter.unmarshall(response);
+                subscriptions.push(item);
+            });
+            if (queryResponse.LastEvaluatedKey) {
+                // To allow subscriptionQuery.ExclusiveStartKey = queryResponse.LastEvaluatedKey,
+                // as otherwise there is an error with syntax
+                Object.defineProperty(subscriptionQuery, 'ExclusiveStartKey', {
+                    value: queryResponse.LastEvaluatedKey,
+                });
+            }
```
carvantes: you should use the correct types so that there's no need of `defineProperty` workarounds.

here `subscriptionQuery` has the type of the return type of `DynamoDbParamBuilder.buildGetActiveSubscriptions()`(specifically the type is `{ IndexName: string; TableName: string; ExpressionAttributeNames: { "#subscriptionStatus": string }; KeyConditionExpression: string; ExpressionAttributeValues: DynamoDB.AttributeMap }`) so TS does not allow to simply add the `ExclusiveStartKey` property to it.

A simple solution is to just use `const subscriptionQuery: any = ...`. It is recommended to avoid using `any` but for this short lived variable it may be ok.

A better solution would be to use a type that correctly describes a DDB query. Looks like the sdk client exposes the type `QueryInput`. Ideally functions like `DynamoDbParamBuilder.buildGetActiveSubscriptions` should use that return type.

carvantes (changes requested):

src/dataServices/dynamoDbDataService.ts
ssvegaraju: I updated the buildGetActiveSubscriptions to return type QueryInput, as that seemed much cleaner as you suggested!

src/dataServices/dynamoDbDataService.test.ts
ssvegaraju: Thanks for catching that

carvantes (approved):

PR #144, opened 2022-03-21
feat: add Batch Bundle support

ssvegaraju:
Issue #, if available:

Description of changes: 
Implemented the processBatch method. To test, I deployed to an AWS Account and made sure all of the different CRUD operations completed successfully.

As for testing, Two points remain that will be caught in a different PR with integration tests: Load testing (very large bundles) and api throttling. In addition, this PR is the first in a series of PRs:
1) Release this PR to npm
2) Update persistence package on the interface and release that to npm
3) Update the interface package on the routing package and release that to npm
4) Update the routing package on the interface package once again and release to npm,
5) Update the interface package on the deployment package.


src/dataServices/dynamoDbBundleService.ts
```diff
@@ -93,10 +93,38 @@ export class DynamoDbBundleService implements Bundle {
...
+                success: true,
+                message: 'Successfully processed bundle',
+                batchReadWriteResponses,
+            };
+        } catch (e) {
+            logger.info('Failed to process batch items', e);
+            return {
+                success: false,
+                message: `failed to process bundle ${e}`,
+                batchReadWriteResponses: [],
```
Bingjiling: When there's failure, the response for each item should still be sent back, instead of sending an empty response.

src/dataServices/dynamoDbBundleServiceHelper.test.ts
```diff
@@ -290,3 +296,160 @@ describe('populateBundleEntryResponseWithReadResult', () => {
...
+                id: 'write',
+            },
+        ];
+
+        sinon
+            .stub(DynamoDbHelper.prototype, 'getMostRecentUserReadableResource')
+            .withArgs('Patient', 'read', undefined)
+            .returns(Promise.resolve(readResource));
+
+        const actualResponse = await DynamoDbBundleServiceHelper.sortBatchRequests(
```
Bingjiling: Is the order of the response same as request if they were sorted first? http://hl7.org/fhir/R4/http.html#transaction-response

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -253,4 +256,231 @@ export default class DynamoDbBundleServiceHelper {
...
+                    vid += 1;
+                    item = DynamoDbUtil.prepItemForDdbInsert(
+                        { ...item?.resource, ...request.resource },
+                        id,
+                        vid,
+                        DOCUMENT_STATUS.AVAILABLE,
+                        tenantId,
+                    );
+                    // we need to delete the old verison, and create a new version
+                    updateRequests.push([
```
Bingjiling: What happens if the update on old resource goes through and the new resource fail to create?

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -253,4 +256,211 @@ export default class DynamoDbBundleServiceHelper {
...
+            // eslint-disable-next-line no-await-in-loop
+            const batchExecuteResponse = await dynamoDb
+                .batchExecuteStatement({
+                    Statements: [...statements],
+                })
+                .promise();
+            batchExecuteResponse?.Responses?.forEach((response, index) => {
+                if (response.Error) {
+                    console.log('Unable to process request: ', response.Error);
+                    const indexOfFailure = batchReadWriteResponses.findIndex((x) => x.id === batch[index].id);
```
Bingjiling: This wouldn't work if there're two requests regarding the same resource. Maybe add the original index as part of the editRequests so it can always map back.

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -253,4 +256,211 @@ export default class DynamoDbBundleServiceHelper {
...
+            const statements = batch.map((x) => {
+                return { Statement: x.Statement };
+            });
+            // eslint-disable-next-line no-await-in-loop
+            const batchExecuteResponse = await dynamoDb
+                .batchExecuteStatement({
+                    Statements: [...statements],
+                })
+                .promise();
+            batchExecuteResponse?.Responses?.forEach((response, index) => {
```
Bingjiling: When would the Response be empty? Is it when there's no write requests?

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -253,4 +256,211 @@ export default class DynamoDbBundleServiceHelper {
...
+                    }
+                }
+            });
+        }
+        return updatedResponses;
+    }
+
+    // ExecuteStatement requires objects to be in the form a PartiQL tuple
+    // this means strings must be single quotes
+    static convertToPartiQLString(value: any) {
```
Bingjiling: This looks hard to read and maintain, is there a library or something that does this? If using this function is unavoidable for inserting with batchExecuteStatement method, maybe we should switch back to the batch writes item method, which does not require parsing the object like this.

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -253,4 +256,211 @@ export default class DynamoDbBundleServiceHelper {
...
+                        vid: item.meta.versionId,
+                        operation: request.operation,
+                        lastModified: item.meta.lastUpdated,
+                        resourceType,
+                        resource: item,
+                    });
+                    break;
+                }
+                case 'update': {
+                    if (vid === 0) {
```
Bingjiling: Does vid === 0 means it's deleted? If so, an error result should still be returned for that request.

src/dataServices/dynamoDbBundleServiceHelper.ts
ssvegaraju: I think this was a holdover from trying to solve the issue with async/await in a try/catch, the `batchExecuteResponse` will never be empty in this scenario.

src/dataServices/dynamoDbBundleServiceHelper.ts
ssvegaraju: I did search for a library to do this, but couldn't find anything through the AWS docs or any other libraries. I will switch back to batchWriteItem for both creates and updates 👍

src/dataServices/dynamoDbBundleServiceHelper.ts
ssvegaraju: vid === 0 means that the resource was not found, and that we already appended the error to `batchReadWriteResponses`, so we don't have to continue there. Unfortunately using `continue` or `break` in the earlier if statement caused unexpected behavior with regards to the asynchronous behavior of the for loop.

src/dataServices/dynamoDbBundleServiceHelper.ts
ssvegaraju: That makes more sense, I'll do that instead!

src/dataServices/dynamoDbBundleServiceHelper.ts
Bingjiling: Emmm so what would happen if the question marks are removed?

ssvegaraju: I forgot to remove the question mark from after `batchExecuteResponse` 😅. The one after `Responses` is just to make sure the array isn't empty (in the case we have no requests).

ssvegaraju: Oh wait nvm this snippet is out of date. The question mark is gone in the PR

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -253,4 +256,215 @@ export default class DynamoDbBundleServiceHelper {
...
+                    Statements: [...statements],
+                })
+                .promise();
+            batchExecuteResponse.Responses?.forEach((response, index) => {
+                if (response.Error) {
+                    console.log('Unable to process request: ', response.Error);
+                    // get the position of the batch element at index in the larger batchReadWriteResponses array
+                    updatedResponses[batch[index].index] = {
+                        ...batchReadWriteResponses[batch[index].index],
+                        vid: '', // indicate the request failed
```
Bingjiling: It's a good habit to name field what they're used for, instead of using another field to indicate. Here it'll be clearer to have an `error` field.

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -253,4 +256,215 @@ export default class DynamoDbBundleServiceHelper {
...
+                        [RESOURCE_TABLE]: [...statements],
+                    },
+                })
+                .promise();
+            batchExecuteResponse.UnprocessedItems?.[RESOURCE_TABLE].forEach((item, index) => {
+                console.log('Unable to process request: ', item);
+                // get the position of the batch element at index in the larger batchReadWriteResponses array
+                updatedResponses[batch[index].index] = {
+                    ...batchReadWriteResponses[batch[index].index],
+                    vid: '', // indicate the request failed
```
Bingjiling: Same as line 463 - Create `error` filed here as well

src/dataServices/dynamoDbBundleServiceHelper.ts
Bingjiling: The unexpected behavior may came from the index+1 not being ran at the end? We can look at it together if it helps. Using vid to indicate if there's error in the read operation is confusing.

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -253,4 +256,215 @@ export default class DynamoDbBundleServiceHelper {
...
+                        lastModified: '',
+                        resource: item?.resource,
+                        resourceType: request.resourceType,
+                    });
+                    break;
+                }
+                default:
+                    break;
+            }
+            index += 1;
```
Bingjiling: Name it something like `originalRequestIndex`? Using just `index` is confusing

src/dataServices/dynamoDbBundleServiceHelper.test.ts
```diff
@@ -290,3 +297,279 @@ describe('populateBundleEntryResponseWithReadResult', () => {
         expect(actualResult).toEqual(expectedResult);
     });
 });
+
+describe('processBatchRequests', () => {
```
Bingjiling, line 301: Add a test to deliberately check the errors were mapped back to the correct item? e.g. Two success and two fail in one request and the errors were mapped back correctly.

Bingjiling (commented):
Looks good! Add one more test to deliberately check the errors were mapped back to the correct item, and we're good to go.

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -253,4 +256,207 @@ export default class DynamoDbBundleServiceHelper {
...
+                        vid,
+                        DOCUMENT_STATUS.AVAILABLE,
+                        tenantId,
+                    );
+
+                    createRequests.push({
+                        PutRequest: {
+                            Item: DynamoDBConverter.marshall(item),
+                        },
+                        index: originalRequestIndex,
```
Bingjiling: Nit pick: use originalRequestIndex as key directly, so in other methods it's clear what this field is for as well.

Bingjiling (dismissed):

src/dataServices/dynamoDbBundleServiceHelper.ts
ssvegaraju: done!

Bingjiling (approved):

Issue #152, opened 2022-08-05
[Bug] Back-to-back batch updates of the same record creates duplicate records


DJK-Cognosante:
**Describe the bug**
Sending a batch update with a PUT to a specific practitioner, and then _after_ the first batch resolves and returns a response, immediately sending another batch update containing a PUT to that same specific practitioner, creates a duplicate record rather than updating. 

- Both records use the same ID, and only the latest version is returned when calling GET against that practitioner.
- When running a search based on e.g. name or NPI, both versions of the record are returned.
- Calling DELETE against the record will only delete the latest version, but the previous version is left ACTIVE in DynamoDB, and therefore still returns in the search results.
- Calling DELETE again to try and delete the still-active record returns 404 Not Found.

**To Reproduce**
1. Create a new practitioner record with a unique name.
2. Send a batch request to fhirworks containing a PUT request to `/Practitioner/${id of record you just created}`
3. As soon as that request resolves, send it again. You have to send it _after_ the previous request resolves, but before DynamoDB's commit window so the changes aren't quite persisted to the DB or the search service yet (admittedly I don't know much about how it works under the hood; this is just my understanding based on preliminary poking around).
4. Perform a search against the practitioner endpoint based around the practitioner's name.
5. You will see two records in the search results with the same ID.
6. Send a DELETE request against the practitioner and re-run the search.
7. You will now see one record in the search results.
8. Send a GET request to the practitioner endpoint, using the ID of the practitioner.
9. You will get 404 Not Found, despite that practitioner appearing in the search results.

**Expected behavior**
Successive PUT requests to the same endpoint (batched or not) should be queued to run sequentially, and await the DynamoDB and/or Elasticsearch commit time to ensure duplicate records are not created.

**Versions (please complete the following information):**
 - Release Version installed: 4.3.1

**Additional context**
We are working on a bulk import process, and our sample file contained the same ~50 records repeated over and over again just to quickly fill out test data. The records were grouped up into chunks of 10-20 and then imported, making sure that each individual request contains no duplicates, and that each request is awaited and fully resolved before doing the next one. We now have 18000+ duplicate records in our test system that we cannot delete by any means.

For the time being, as a workaround, we've had to simply state that all import files must be manually checked for duplicate identifiers before being imported, to make 100% sure that the same record isn't updated twice in too short of a timespan. This obviously isn't ideal as some of the files contain millions of records.

DJK-Cognosante:
[...]

PR #159, opened 2022-12-09
Bug: Fix Batch Upsert

kcadette:
**Describe the bug**
when [Update As Create ](https://www.hl7.org/fhir/http.html#upsert)is supported, creating a resource using PUT, batch bundles attempt to only find the resource first instead of creating them as they should. 

**Expected behavior**
Bundle should create the resources in the bundle.
Issue #, if available:

Description of changes:
Added a check for if update create is supported and whether the operation is update. If so, and the resource does not exist we we will create the resource. 


ssvegaraju (commented):
This looks good to me! Couple things: 
- Could we rename variables like `ddbHelper2` to something a bit more meaningful?
- Were you able to test this on a deployed account?

src/dataServices/dynamoDbBundleServiceHelper.ts
```diff
@@ -257,10 +258,41 @@ export default class DynamoDbBundleServiceHelper {
         return { stagingResponse, itemLocked };
     }
 
+    private static async createBatchResource(createObject: any) {
```
nguyen102: This function has a [side effect](https://thejs.dev/jmitchell/what-are-side-effects-and-what-you-can-do-about-them-jws) of modifying the variables `createRequests` and `batchReadWriteResponses`. While this works, it's dangerous because it's not immediately clear that this function will be changing the state of variables from the parent function. 

I would recommend using this function to generate the individual requested needed for `createRequests` and `batchReadWriteResponses`. Return the individual request and let the parent function add the individual request to the appropriate arrays. 

An example is shown [here](https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blame/0d245c2db2ba8629c6ac077746753431e1edfc0d/src/dataServices/dynamoDbBundleServiceHelper.ts#L67)

src/dataServices/dynamoDbBundleServiceHelper.test.ts
```diff
@@ -602,14 +602,109 @@ describe('sortBatchRequests', () => {
...
+        sinon.stub(ddbHelper2, 'getMostRecentUserReadableResource').callsFake(function stubbedGet() {
+            throw new ResourceNotFoundError('Patient', 'read');
+        });
+
         const actualResponse = await DynamoDbBundleServiceHelper.sortBatchRequests(operations, ddbHelper);
         expect(actualResponse.batchReadWriteResponses).toMatchObject(expectedBatchReadWriteResponses);
         expect(actualResponse.deleteRequests).toMatchObject(expectedDeleteRequests);
         expect(actualResponse.writeRequests).toMatchObject([...expectedCreateRequests, ...expectedUpdateRequests]);
+
+        const updateCreateSupportedResponse = await DynamoDbBundleServiceHelper.sortBatchRequests(
```
nguyen102: Hmm, should we just move this test logic to another test case. The test case would be specifically for testing the `updateCreateSupported = true` case. The original test can be for the test case `updateCreateSupported = false`.

kcadette: fixed

awsdrbalaji:
For testing, have we considered combinatorial testing here. Going forward we should consider that for any new test we add here and explain why it is not applicable if we are not adding them. This is coming from the COE.

src/dataServices/dynamoDbBundleServiceHelper.test.ts
```diff
@@ -484,75 +534,210 @@ describe('sortBatchRequests', () => {
...
+                originalRequestIndex: 3,
             },
         ];
-
-        const expectedBatchReadWriteResponses: BatchReadWriteResponse[] = [
+        const expectedBatchReadWriteResponsesUpsert: BatchReadWriteResponse[] = [
             {
                 id: readResource.resource.id,
-                vid: '1',
+                vid: '0',
```
ssvegaraju, line 712: The expected vid for a created resource should be 1. Is there a reason why these tests have all been changed to throw 404 Not Found, instead of creating new tests for the upsert? I think we should avoid changes to existing tests as much as possible so we can ensure there aren't breaking changes.

yarn.lock
```diff
@@ -6084,4 +6084,4 @@ yargs@^16.0.0, yargs@^16.2.0:
 yocto-queue@^0.1.0:
   version "0.1.0"
   resolved "https://registry.yarnpkg.com/yocto-queue/-/yocto-queue-0.1.0.tgz#0294eb3dee05028d31ee1a5fa2c556a6aaf10a1b"
-  integrity sha512-rVksvsnNCdJ/ohGc6xgPwyN8eheCxsiLM8mxuE/t/mOVqJewPuO1miLpTHQiRgTKCLexL4MeAFVagts7HmNZ2Q==
+  integrity sha512-rVksvsnNCdJ/ohGc6xgPwyN8eheCxsiLM8mxuE/t/mOVqJewPuO1miLpTHQiRgTKCLexL4MeAFVagts7HmNZ2Q==
```
ssvegaraju: Was this package used in this PR? I didn't see it, may want to remove this change if it is unused.

ssvegaraju (commented):
Were you able to run the [integ-tests for batch](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/integration-tests/batchBundle.test.ts) btw?

src/dataServices/dynamoDbBundleServiceHelper.test.ts
kcadette: This is the expected response for operation: read, on a resource that doesn't exists
All tests with 404 Not Found are added tests - the original tests are still there. I added them to the existing tests for when the resource doesn't exist. In such cases: When  updateCreateSupported=false, there should be some 404 responses when the resource is not found (on read, delete, and update). When updateCreateSupported=true, there should be 404 responseswhen the resource is not found on read and delete.

ssvegaraju: Gotcha I got confused by the git addition/removals 😅

ssvegaraju (approved):

russellott:
Hey all  - we pulled this refined code and did some testing:

While we can verify that it does enable “Upsert” processing for Bundles of type = batch, it has a pretty serious bug in that, when the update is creating a new version of an existing resource, it leaves the “documentStatus” of the old version of the resource as “AVAILABLE” instead of setting it to “DELETED”.

It’s possible this may be related to this other existing issue, but I’m not sure: https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/issues/152

Has anyone else tested and experienced the same?

kcadette:
@russellott we're looking into this.

PR #160, opened 2022-12-22
[WIP] fix: check whether vid is a number

kcadette:
Issue #, if available:

Description of changes:
Update http error code to BadRequest when incorrect input is used for version id.
I deployed and tested this change in a development acct

### DIFF
--- a/src/dataServices/dynamoDbDataService.ts
+++ b/src/dataServices/dynamoDbDataService.ts
@@ -19,7 +19,7 @@ import {
     GetExportStatusResponse,
     InitiateExportRequest,
     InvalidResourceError,
-    InvalidSearchParameterError,
+    BadRequestError,
     isResourceNotFoundError,
     PatchResourceRequest,
     Persistence,
@@ -110,7 +110,7 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
         const { resourceType, id, vid, tenantId } = request;
         const vidNumber = parseInt(vid, 10);
         if (!vidNumber) {
-            throw new InvalidSearchParameterError('Invalid versionId');
+            throw new BadRequestError('Invalid versionId');
         }
         const params = DynamoDbParamBuilder.buildGetItemParam(id, vidNumber, tenantId);
         const result = await this.dynamoDb.getItem(params).promise();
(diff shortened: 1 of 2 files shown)


src/dataServices/dynamoDbDataService.ts
```diff
@@ -107,6 +107,9 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
     async vReadResource(request: vReadResourceRequest): Promise<GenericResponse> {
         this.assertValidTenancyMode(request.tenantId);
         const { resourceType, id, vid, tenantId } = request;
+        if (Number.isNaN(Number(vid))) {
```
sevu1: Why not use the same code as below? `parseInt(vid, 10)`
```
const version = `parseInt(vid, 10)`;
if (!version).... 
```

src/dataServices/dynamoDbDataService.ts
```diff
@@ -107,6 +107,9 @@ export class DynamoDbDataService implements Persistence, BulkDataAccess {
...
+            throw new UnauthorizedError('Invalid versionId');
```
sevu1: Is it unauthorized or bad request?

sevu1:
Also might be a good candidate for unit test

src/dataServices/dynamoDbDataService.ts
sevu1: One thing to consider: this code won't work for 0..0.99, e.g. 
image

kcadette: makes sense, I updated this logic

src/dataServices/dynamoDbDataService.ts
kcadette: From review we want unauthorized for this sort of request

sevu1: But description says it should be 400 - Bad request (which makes sense); Unauthorized is 403. If `throw new UnauthorizedError` is mapped to 400, this is something we should address

sevu1: Apparently it is 403: https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/router/routes/errorHandling.ts#L35 .
Either we should add new error class to be mapped to 400 or rename `isInvalidSearchParameterError` to be more generic - this might be a bigger footprint on number of changed files, so might be former is better

ssvegaraju (commented):
You will need to update the version number of the fhir-works-on-aws-interface package in `package.json` and run `yarn install` to fix the unit test issues once the interface changes are released, but other than that the PR Looks good!

